PR #80, opened 2020-03-12
Change invoice style and creation QR

brick9527:
Changed

- style of invoice page

Added

- some functions about creating QR code

I didn't change the page structure. Just completing and adjusting the page.
image

### DIFF
--- a/src/views/forms/FormTransferTransaction/FormTransferTransactionTs.ts
+++ b/src/views/forms/FormTransferTransaction/FormTransferTransactionTs.ts
@@ -34,7 +34,7 @@ import {TransactionFactory} from '@/core/transactions/TransactionFactory'
 import {AddressValidator} from '@/core/validation/validators'
 import {MosaicInputsManager} from './MosaicInputsManager'
 import {MosaicService} from '@/services/MosaicService'
-import {ITransaction} from '@/views/pages/dashboard/invoice/DashboardInvoicePageTs'
+import {ITransactionEntry} from '@/views/pages/dashboard/invoice/DashboardInvoicePageTs'
 
 // child components
 import {ValidationObserver} from 'vee-validate'
@@ -423,11 +423,11 @@ export class FormTransferTransactionTs extends FormTransactionBase {
   triggerChange() {
     if (this.formItems.recipientRaw && this.formItems.recipientRaw !== '') {
       const transactions = this.getTransactions()
-      const data: ITransaction[] = []
+      const data: ITransactionEntry[] = []
       transactions.map((item: TransferTransaction) => {
         data.push({
           transaction: item,
-          mosaicAttachment: this.mosaicsToAttachments(item.mosaics),
+          attachments: this.mosaicsToAttachments(item.mosaics),
         })
       })
 
--- a/src/views/forms/FormTransferTransaction/FormTransferTransactionTs.ts
+++ b/src/views/forms/FormTransferTransaction/FormTransferTransactionTs.ts
@@ -34,14 +34,7 @@ import {TransactionFactory} from '@/core/transactions/TransactionFactory'
 import {AddressValidator} from '@/core/validation/validators'
 import {MosaicInputsManager} from './MosaicInputsManager'
 import {MosaicService} from '@/services/MosaicService'
-
-interface MosaicAttachment {
-  mosaicHex: string
-  /**
-   * Relative amount
-   */
-  amount: number
-}
+import {ITransaction} from '@/views/pages/dashboard/invoice/DashboardInvoicePageTs'
 
 // child components
 import {ValidationObserver} from 'vee-validate'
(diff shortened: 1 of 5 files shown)


src/views/pages/dashboard/invoice/DashboardInvoicePageTs.ts
```diff
@@ -70,7 +47,9 @@ export interface BalanceEntry {
   subscriptions() {
     const qrCode$ = this
       .$watchAsObservable('transactionQR', {immediate: true})
-      .pipe(pluck('newValue'),
+      .pipe(
+        debounceTime(500),
```
decentraliser: try throttleTime

brick9527: The difference between `throttleTime` and `debounceTime` :

When the value is 500(ms).

- throttleTime

It will call this function once every 500ms, no matter the value changed or not.

- debounceTime

When the value of `transactionQR` changed, it will call this function after 500ms, no matter the value changed how many times in next 500ms.

When user inputing the form, the transaction will change with user inputing. So the value of `transactionQR` will change a lot times, but we just need the last transaction to create the QR. For saving the times of computing, I think `debounceTime` is better.

I tried `throttleTime`, I feel not good.

decentraliser: did you put the leading and trailing options? 
 throttleTime(EVENTS_THROTTLING_TIME, asyncScheduler, {leading: true, trailing: true}),

brick9527: No, I just `throttleTime(500)`.

src/views/pages/dashboard/invoice/DashboardInvoicePageTs.ts
```diff
@@ -150,14 +128,45 @@ export class DashboardInvoicePageTs extends Vue {
...
+    }
+  }
+
+  /**
+   * The first transaction in the list stack
+   * @type {ITransaction}
+   */
+  get shiftTransaction(): ITransaction {
+    if (this.transactions.length) {
+      return this.transactions.shift()
```
evias: if you call this method 2 times, it will return different results but it should not. Please use `this.transactions[0]` instead.

src/views/pages/dashboard/invoice/DashboardInvoicePageTs.ts
```diff
@@ -150,14 +128,45 @@ export class DashboardInvoicePageTs extends Vue {
...
+    if (this.shiftTransaction) {
+      return this.shiftTransaction.transaction as TransferTransaction
+    }
+  }
+
+  /**
+   * The first transaction in the list stack
+   * @type {ITransaction}
+   */
+  get shiftTransaction(): ITransaction {
```
evias: rename to `transactionEntry`

src/views/pages/dashboard/invoice/DashboardInvoicePageTs.ts
```diff
@@ -29,33 +30,9 @@ import FormTransferTransaction from '@/views/forms/FormTransferTransaction/FormT
...
-   * Mosaic name
-   * @type {string}
-   */
-  name: string
-  /**
-   * Relative amount
-   * @type {number}
-   */
-  amount: number
+export interface ITransaction {
```
evias: rename to `ITransactionEntry`

src/views/pages/dashboard/invoice/DashboardInvoicePageTs.ts
```diff
@@ -29,33 +30,9 @@ import FormTransferTransaction from '@/views/forms/FormTransferTransaction/FormT
...
+  transaction: TransferTransaction
+  mosaicAttachment: MosaicAttachmentType[]
```
evias: rename field to `attachments` (and should be plural)

NemTechCN: ok,thanks

src/views/pages/dashboard/invoice/DashboardInvoicePageTs.ts
decentraliser: ok, please try throttleTime(500, asyncScheduler, {leading: true, trailing: true})

import {asyncScheduler} from 'rxjs'

throttle time will not call this function every 500 ms

with this options, it calls the function as soon as the form will be changed (debounceTime will wait for 500ms), and the next call will be the one triggered by the last change that occurred from the first call to 500ms after the first call, which is why it is better in terms of UX

https://rxmarbles.com/#debounceTime
https://rxmarbles.com/#throttleTime

But tbh, I think we could just not throttle nor debounce

decentraliser: debouncing and throttling should be done for expensive operations, this one is not expensive

brick9527: Finally, I removed it.
But we must be attention to the users whoes computer is not good. Maybe creating QR so frequently is not good.

src/views/forms/FormTransferTransaction/FormTransferTransactionTs.ts
```diff
@@ -66,12 +59,20 @@ import MaxFeeAndSubmit from '@/components/MaxFeeAndSubmit/MaxFeeAndSubmit.vue'
...
+  amount: number
+}
+
+export interface MosaicAttachmentType {
+  id: MosaicId
+  mosaicHex: string
+  name: string
+  amount: number
+  uid: number
 }
```
evias: duplicate interfaces ... the second interface just has more fields, these can be made optional in first interface and then only use that one.

brick9527: resolved

PR #121, opened 2020-03-18
Account mnemonic import component v1

David0817:
import mnemonic component
- Input box style
- add copy and paste


src/components/MnemonicInput/MnemonicInputTs.ts
```diff
@@ -0,0 +1,118 @@
...
+      this.inputWord = '';
+      this.initInput();
+    } else {
+      //control the keyboard input rules
+      this.inputWord = newVal.replace(/[^a-zA-Z]/g, '')
+      //determine if the input is editing status
+      if (!this.isEditing && !!this.inputWord) {
+        this.isEditing = true
+      }
+    }
```
evias: instead of a **watch**, it is preferred to use computed properties.

```
public  get userInput(): string {
    return this.inputWord || ''
}

public set userInput(input: string) {
    // do modifications/split here 
}
```

evias:
rebased and merged in master with #c66ef763

PR #132, opened 2020-03-19
Repository factory

fboucquez:
Hi guys,

This PR is a small refactoring that uses the SDK's repository factory to create the HTTP clients and listeners instead of the RESTService. 

This PR also includes https://github.com/nemfoundation/symbol-desktop-wallet/pull/131 I need to keep  the testing going


src/services/RESTService.ts
```diff
@@ -191,18 +111,32 @@ export class RESTService extends AbstractService {
     ]}
   }
 
+  /**
+   * It creates the RepositoryFactory used to build the http repository/clients and listeners.
+   * @param url the url.
+   */
+  public static createRepositoryFactory(url: string): RepositoryFactory {
+    //TODO Extend RepositoryFactoryHttp in tha TS SDK to allow ws:// kind of urls.
+    //TODO Why the WebSocket as webSocketInjected? Can we avoid this?
```
decentraliser: no, this is necessary when running in the browser

evias: @fboucquez client builds use the browser's `WebSocket` class. Would require some testing to remove this but since the wallet is *always* browser served, I believe it could be removed (please mind the testing with _artifacts_ too)

src/store/Mosaic.ts
```diff
@@ -164,12 +172,12 @@ export default {
         knownMosaics: withFeed.mosaics
       }
     },
-    async INITIALIZE_FROM_NEMESIS({commit, dispatch}, nodeUrl) {
+    async INITIALIZE_FROM_NEMESIS({commit, dispatch}, repositoryFactory: RepositoryFactory) {
       // read first network block to identify currency mosaic
 
-      dispatch('diagnostic/ADD_DEBUG', 'Store action mosaic/INITIALIZE_FROM_NEMESIS dispatched with nodeUrl: ' + nodeUrl, {root: true})
+      dispatch('diagnostic/ADD_DEBUG', 'Store action mosaic/INITIALIZE_FROM_NEMESIS dispatched with nodeUrl: ' + repositoryFactory, {root: true})
```
evias: think this log will error out (or print `[Object...]`)

src/services/RESTService.ts
fboucquez: The TODOs are more for the SDK than for the wallet.  I'll find the way to configure the RepositoryFactory so you can tell which WebSocket you want to use and if the URL is HTTP: or ws:

src/store/Mosaic.ts
fboucquez: Hi Greg, I've kind of fixed the log line by removing the URL. I think the whole how to load configuration from rest/db/database can be improved. This logline will change once we do the improvement. I can POC the improvement in the next days if you are OK.

evias: this log was working fine ; my point is that now it will be adding wrong information in the debug console and would not display the node anymore.

Best would be to replace nodeUrl with repositoryFactory.url (not sure if there is such property but i guess so)

evias: Wrt to rest/db im glad to see suggestions yes :+1:

fboucquez: I added the nodeUrl value back to the log. repositoryFactory doesn't know that it's talking to a rest url. repositoryFactory.url not possible atm.

PR #137, opened 2020-03-20
fix generate wallet finish button padding

### LINKED ISSUE #133: When generating the wallet completion step page, the button is blocked
The button is not fully displayed,When the width is greater than 2 times the height.
like this:
Button is blocked

David0817:
button padding fixed #133


evias:
merged with 861ae5f5

Issue #151, opened 2020-03-25
close button of the asetslist is invalid，when the list is larger than 10 and is pulled to the bottom


David0817:
<img width="271" alt="close button of the wallet list is invalid" src="https://user-images.githubusercontent.com/18167466/77505886-5a2d6000-6e9f-11ea-8d64-c39e0acacd6d.png">

evias:
merged with 28e1ac6c

PR #155, opened 2020-03-25
fix Asetslist close button #151

David0817:
Fix close button failure when list is pulled down to the bottom #151


src/components/MosaicBalanceList/MosaicBalanceList.less
```diff
@@ -37,7 +37,7 @@
...
+  background-image: url("../../views/resources/img/monitor/mosaics/unselected.png");
```
evias: thanks. Much better.

PR #177, opened 2020-03-31
Database and storage refactoring

fboucquez:
Hi team,

I've been revisiting the wallet for the last week and I have seen a few points that I think they can be improved on. 

If you have the time, please revisit it, I think there are valid points to be considered. I recommend checking out the code and running it locally, not just revisit the changed code in GitHub.

## Data/Storage:

1) Models without typed attributes:  Components and the compiler cannot tell the attributes and the types of these values. Clients get the values by key and then they need to downcast to use them or ignore the types. Time is lost looking for the right attribute names.  Typos (invalid attribute name) and types (invalid operation on a value) errors are found at runtime.
    - You cannot tell the known attributes of Models like [MosiacModel](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/core/database/entities/MosaicsModel.ts) by just looking at it. Clients need to get a value using strings (e.g. [Mosaic](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/store/Mosaic.ts#L188)).
    - The name of the attributes are in table objects like [MosaicsTable](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/core/database/entities/MosaicsTable.ts#L26). Devs need to refer to this table which is time-consuming and error-prone.
    - This problem can be reduced by converting the model into another object like in the [NamespaceModel.object](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/core/database/entities/NamespacesModel.ts#L53). But this is another mapper returning another object with an anonymous type. Anonymous type is OK for local operations but when they are propagated in other parts of the code it's better to have a type name.

2) The Storage framework can be simplified: For a model object like Wallet you need  [WalletModel](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/core/database/entities/WalletsModel.ts), [WalletTable](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/core/database/entities/WalletsTable.ts), [WalletRepository](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/repositories/WalletsRepository.ts), etc. I understand the idea of having a more _robust_ database but I don't think it's necessary in the Wallet Case. The data stored is either cache data (Network, Namespace, Mosaic) or almost-single-tenant configuration (Settings, Wallet, Account).

3) There is duplicated and boilerplate code: [WalletsModel](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/core/database/entities/WalletsModel.ts#L53) vs [Namespaces](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/core/database/entities/NamespacesModel.ts#L35) - [WalletRepository](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/repositories/WalletsRepository.ts#L66) vs [NamespaceRepository](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/repositories/NamespacesRepository.ts#L66)

###  Data/Storage improvements in the PR:

1) Migrate the storage object like Model and Table to just a simple POJO (not methods, just simple attributes) that can be serialized/deserialized to JSON. Examples are [NamespaceModel](https://github.com/NEMStudios/symbol-desktop-wallet/blob/storage-refactoring/src/core/database/entities/NamespaceModel.ts#L30), [MosaicModel](https://github.com/NEMStudios/symbol-desktop-wallet/blob/storage-refactoring/src/core/database/entities/MosaicModel.ts). The JSON schema as been kept for Account and Mosaic tables. No migration is required, they should when starting the new version. The Settings table has been slightly modified to fix the snake_case toward camelCase (example default_fee to [defaultFee](https://github.com/NEMStudios/symbol-desktop-wallet/blob/storage-refactoring/src/core/database/entities/SettingsModel.ts#L26)). The Table and Repository objects are not needed anymore and removed. Old storage classes have been also removed (they are kept in git history for reference).

2) The storage has been simplified to a generic object that can store/retrieve/delete an object of a given type [SimpleObjectStorage](https://github.com/NEMStudios/symbol-desktop-wallet/blob/storage-refactoring/src/core/database/backends/SimpleObjectStorage.ts#L30)

3) There is no storage related code duplicated for each table/model.

### Data/Storage PR TODO:

The migration feature of the new storage hasn't been created. My plan is to add `{version:x , data: RealStoredData}` payload in the tables. This requires changing the table schemas a little bit. The version would be kept generic at the top level and not per row/entity inside the table.  Example: there is no need to keep a version per Account, all the stored accounts in the same table will have the same top-level version. In the case of migration, you migrate the whole lot. It shouldn't be possible to have an account of one version with an account of another version at the same time, everything should have the same version, or migrated from the same version

## Modules separation of concern

Another issue I notice it's the high coupling between the different modules of the application. UI components can talk to service, store, repositories, etc. Stores may talk to repositories, service, etc. Services are aware of the vuex state.  Examples [here](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/views/forms/FormTransferTransaction/FormTransferTransactionTs.ts#L205), [here](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/views/pages/accounts/import-account/wallet-selection/WalletSelectionTs.ts#L139), [here](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/services/MosaicService.ts#L256), [here] (https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/services/MosaicService.ts#L408)

If a function uses `getter('someState')` style, it may not be refreshed then when `'someState'` is changed.

My recommendation is to define the dependencies a bit better: One example I'm trying to follow is (in general):

- UI Components: only get the status from State and getters: They can only announce operations via actions to the Stores.  
- Services are not aware of Vue. They can only talk to rest and the db storage for caching or configuration. if they don't know the `$store`, they will be simpler to unit test. Getting values using $store is a bad shortcut. as they are not well-typed and it's hard to tell from method contract level the values the service needs to perform an operation.
- Store is the glue between the Components and Services. They receive actions from UI components, perform an operation via services and change their state. Then the UI gets updated when the State is changed.

The best improvement example I've done is around [Mosaic](https://github.com/NEMStudios/symbol-desktop-wallet/blob/storage-refactoring/src/store/Mosaic.ts) and [Store](https://github.com/NEMStudios/symbol-desktop-wallet/blob/storage-refactoring/src/store/Mosaic.ts) where Components only gest the information via `mapGetters`

## Added features in the PR:
- Known Peers nodes are added to the peer selector automatically after querying rest getNodePeers[](https://nemtech.github.io/symbol-openapi/#operation/getNodePeers) https://github.com/nemfoundation/symbol-desktop-wallet/issues/139
- The Peer selector has a built-in  quick filter input (the list is pretty large now)
- The NetworkService and Network storage pings the known list to select the best available peer before creating the network repository. I have seen cases where the default peer is  banned
- There is a database cache for Mosaics and Namespaces. When the entities need to be loaded, the observable first sends to the clients the cached values, then it retrieves and sends the rest values. 1) data is shown if the peer is down, 2) faster user feedback/interaction
- NetworkConfiguration is [loaded](http://api-01.us-west-1.symboldev.network:3000/network/properties), preprocessed and cached. Then the components use the stored configuration to perform the operations https://github.com/nemfoundation/symbol-desktop-wallet/issues/140
- namespaceGracePeriodDuration is now a duration in seconds, not a block height https://github.com/nemfoundation/symbol-desktop-wallet/issues/173
- SDK has been upgraded to `0.17.4-alpha`. I removed an Axios call with an SDK call. No custom rest call is needed.  This means the wallet uses the official open API spec, the official catapult-rest can be used and not the fork
- Simplified the [network.conf.json](https://github.com/NEMStudios/symbol-desktop-wallet/blob/storage-refactoring/config/network.conf.json) file and removed hardcoded testnet only code `networkConfig.networks['testnet-publicTest']` calls like [AboutPage](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/views/pages/settings/AboutPage/AboutPage.vue#L52). If the hardcoded configuration is kept to the minimal, the wallet can be easily ported to different network types or maybe support multiple network types at the same time. Most configuration should be retrieved from rest using just the default for the very initial boot.

## Code improvements in the PR:
- Improved types `veux` State object. Now States like [Mosaic](https://github.com/NEMStudios/symbol-desktop-wallet/blob/storage-refactoring/src/store/Mosaic.ts#L28), [Namespace](https://github.com/NEMStudios/symbol-desktop-wallet/blob/storage-refactoring/src/store/Namespace.ts#L26) have better typed states, getters, mutations, actions, etc. When coding a UI component, and you need a store value, you can easily know the type of the value.
- Simplified action parameters allowing just one type. This avoids having multi-typed params and methods like [here](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/store/Wallet.ts#L76) and [here](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/store/Wallet.ts#L102)
- Improved anonymous types that are used across multiple components. The main example is the new [Signer](https://github.com/NEMStudios/symbol-desktop-wallet/blob/storage-refactoring/src/store/Wallet.ts#L69) type. It was hard to tell the type of `'wallet/currentSigner'`.
- With new Storage Models, the Model can be better reuse in UI Components simplifying the code as the attributes and the types are well known and reducing one-of types and mappers
- Removed `$store` from most services
- Only services talk to the database storage.
- Removed unnecessary downcast/upcast method in TransctionService [method](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/services/TransactionService.ts#L105)
- Incorrect use of WalletType enums. Wallet is using `WalletType.fromDescriptor('Seed')` when it can be simplified with just a `WalletType.SEED`

## General TODOs (future tasks) 
- Keep improving the separation between the different modules.
- `CustomValidationRules` and `ValidationRuleset` are still using the statically configured network properties/configuration. 
- Types in State can still be improved by adding more types to actions. Is it possible to better type the component's mapGetters annotations having a compilation time error if the getter doesn't exist or it has a wrong type?
- When a Components perform an action like [Logout](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/components/PeerSelector/PeerSelectorTs.ts#L245), the UI component is responsible for telling the different stores the actions to be performed increasing coupling. When a new action needs to be performed when logging out, the component needs to be changed adding the dispatch call. I would like to use a more event base model, where one LOGOUT event is fire and different stores subscribe to that Event. One event handler could be the ''diagnostic' store that logs the events into Wallet console reducing the component responsibility to log the action.
- The same listener instance to the same peer can be used in the different Store increasing performance.
- symbol-hd-wallets needs to be upgraded `0.17.4-alpha` so the SDK Account Types are compatible at compilation time without funny [castings](https://github.com/NEMStudios/symbol-desktop-wallet/blob/storage-refactoring/src/services/WalletService.ts#L139).
- Increase unit testing a lot. Not an easy task, although I reckon the PR code would be a bit simpler to unit tests (less coupling).  I would recommend enabling coveralls reports in Travis and GitHub
- Improve code documentation starting from the architectural classes (main, store, services, storage, etc) down to the UI components.
- Table and Transaction views are still using key-value dictionary and maps when better-typed objects could be used.

# Summary:

I know this is a big PR and it has crossed multiple areas. It's not ideal and I do prefer having smaller PR but this is a special, my first take on the Wallet. We shouldn't be doing this. 
Having said that, there are multiple ways of handling this PR.

1) One big PR merge after validating each point, regression testing, and feedback fixing. Then continue the TODOs as individual Issues and PRs
2) Split the PR into several ones a tackle one by one. It may take a lot of time to split. Each PR split may require regression testing that could take a lot of time. How to split is also hard, would we migrate half of the storage keeping both solutions at the same time?
3) Never merge this PR and takes some of the ideas into the code for future improvements and PR.

Happy to discuss each point. 
Cheers

Fernando


decentraliser:
>Types in State can still be improved by adding more types to actions. Is it possible to better type the component's mapGetters annotations having a compilation time error if the getter doesn't exist or it has a wrong type?

There are solutions but they are a bit hacky (but better than nothing IMO). I think that there won't be further fixes before Vue 3 that will be written in TypeScript
[see usage of GetterTree here ](https://codeburst.io/vuex-and-typescript-3427ba78cfa8)

decentraliser:
>Table and Transaction views are still using key-value dictionary and maps when better-typed objects could be used.

About maps, another advantage of not using them is that Vue reactivity engine does not support them. However, Vue 3 will support them
[see Observation Mechanism ](https://medium.com/the-vue-point/plans-for-the-next-iteration-of-vue-js-777ffea6fabf)

config/network.conf.json
```diff
@@ -2,55 +2,29 @@
...
-  "defaultNode": {
-    "protocol": "http",
-    "hostname": "api-01.us-west-1.symboldev.network",
-    "port": 3000,
-    "url": "http://api-01.us-west-1.symboldev.network:3000"
+  "defaultNodeUrl": "http://api-01.ap-northeast-1.symboldev.network:3000",
+  "networkConfigurationDefaults": {
+    "maxMosaicDivisibility": 6,
+    "namespaceGracePeriodDuration": 2592000,
+    "lockedFundsPerAggregate": "10000000",
```
decentraliser, line 9: is it wanted to store this number as a string here?

decentraliser:
Account import seems to be broken

When importing an account from mnemonic and chosing addresses (having the new default address with a balance and a transaction history), the dashboard left section (balance panel and mosaic list) initialize empty, and it throws:
vue.esm.js:1897 TypeError: Cannot read property 'plain' of null
    at e.s (AccountBalancesPanel.vue?4d4a:7)
    at e._render (vue.esm.js:3557)
    at e.i (vue.esm.js:4075)
    at ii.get (vue.esm.js:4488)
    at new ii (vue.esm.js:4477)
    at Rr (vue.esm.js:4082)
    at e.xi.$mount (vue.esm.js:9063)
    at e.xi.$mount (vue.esm.js:11974)
    at init (vue.esm.js:3127)
    at h (vue.esm.js:5989)

in walletChoose, the balances are not showing
After a moment (around 1s), the mosaics and the balance show, but the transaction list is empty

Wallets are not showing in the wallet view

Them when I logout and login, it goes straight to the mnemonic random generation screen (mousemove)

decentraliser:
When starting the app with empty local storage

The language selector is not properly initialized (it shows as "Select" when starting the app with an empty localStorage)
When setting the language in that state, it throws:
vue.esm.js:1897 TypeError: Cannot read property 'accountName' of null
    at v.SET_SETTINGS (AppInfo.ts:128)
    at Array.<anonymous> (vuex.esm.js:747)
    at v.dispatch (vuex.esm.js:438)
    at v.dispatch (vuex.esm.js:332)
    at T.n.dispatch (vuex.esm.js:675)
    at v.SET_LANGUAGE (AppInfo.ts:137)
    at Array.<anonymous> (vuex.esm.js:747)
    at v.dispatch (vuex.esm.js:438)
    at v.dispatch (vuex.esm.js:332)
    at e.set [as language] (LanguageSelectorTs.ts:62)

decentraliser:
Listeners do not seem to work

There are no notifications for new (un)confirmed transactions
password1
When entering in "Create sub namespace" tab, the page stays black, with the error:
vue.esm.js:1897 TypeError: Cannot read property 'length' of undefined
    at fn (CreateSubNamespace.vue?fc2f:1)
    at i (vue.esm.js:2599)
    at e.Ut [as _t] (vue.esm.js:2695)
    at e.o (AssetFormPageWrap.vue?1dd1:1)
    at e._render (vue.esm.js:3557)
    at e.i (vue.esm.js:4075)
    at ii.get (vue.esm.js:4488)
    at new ii (vue.esm.js:4477)
    at Rr (vue.esm.js:4082)
    at e.xi.$mount (vue.esm.js:9063)

decentraliser:
Assets tables

In mosaics table, when clicking "link", the mosaic name does not appear
In namespace table, mosaics do not whow in the dropdown when clicking "link"
In mosaics table, when clicking "link", the namespaces that shows are not the ones of the active wallet

decentraliser:
If I have a bad connectivity, a node switch can let me forever with the loading overlay

decentraliser:
For some reason, the `npm run dev` command is making my browser crash at the first load of localhost

decentraliser:
transaction list 
Sometimes, when switching wallets, the transaction list is not properly initialize
Wallet change can be done from the wallet section, or from the dashboard using the wallet selector

The transaction list sometimes does not clear when switching wallets (when switching to wallet B, I still see transactions of wallet A) untill switching to another view and coming back to the dashboard

src/components/TransactionDetails/TransactionDetailsTs.ts
```diff
@@ -15,40 +15,11 @@
...
-  NamespaceRegistrationTransaction,
-  NetworkType,
-  SecretLockTransaction,
-  SecretProofTransaction,
-  Transaction,
-  TransactionType,
-  TransferTransaction,
-} from 'symbol-sdk'
-
+import {AccountAddressRestrictionTransaction, AccountLinkTransaction, AccountMetadataTransaction, AccountMosaicRestrictionTransaction, AccountOperationRestrictionTransaction, AddressAliasTransaction, AggregateTransaction, HashLockTransaction, MosaicAddressRestrictionTransaction, MosaicAliasTransaction, MosaicDefinitionTransaction, MosaicGlobalRestrictionTransaction, MosaicId, MosaicMetadataTransaction, MosaicSupplyChangeTransaction, MultisigAccountModificationTransaction, NamespaceMetadataTransaction, NamespaceRegistrationTransaction, NetworkType, SecretLockTransaction, SecretProofTransaction, Transaction, TransactionType, TransferTransaction} from 'symbol-sdk'
```
decentraliser, line 18: This should not pass the linter (120 max-len)

It's better to have them in a column so the unused var are easier to spot

fboucquez: We have added prettier with eslint to the ts-sdk. https://github.com/nemtech/symbol-sdk-typescript-javascript/pull/531. Prettier is really good, you may want to consider it for the wallet. I did spend a lot of time trying to replicate the style formatting in Intellij (back in the days in the SDK and last week on the wallet), If I had prettier it would have worked out of the box for any IDE (as it's cli based). It's also integrated with eslint (eslint fails if the code is not 'pretty' enough). Having an IDE independent consistent style is great. eslint --fix only fixes some of the style, not everything (prettier fills the gap). I think prettier can work with Vue, but this needs to be tested out. 

We can integrate prettier after the release and this PR when there aren't too many open PRs

fboucquez:
> transaction list
> Sometimes, when switching wallets, the transaction list is not properly initialize
> Wallet change can be done from the wallet section, or from the dashboard using the wallet selector
> 
> The transaction list sometimes does not clear when switching wallets (when switching to wallet B, I still see transactions of wallet A) until switching to another view and coming back to the dashboard

I haven't revisited the transaction lists. Something that I have noticed is that transactions aren't been cached in db like mosiac or namespaces. Is this expected?  

What I have done for namespaces and mosaic is loading the entities for all known wallets (sdk allows you to load entities for mutliple accounts with one rest call). Then when switching from one wallet to another, the new wallet has the namespaces/mosaics prefetched, feedback to the user is quicker. There will be another loading just to retrieve the latest ones when hitting the page (or maybe in response to another event) I guess we could do something similar with transactions instead of just cleaning up existing and load new ones.

fboucquez:
This PR will need a full round trip of regression testing and fixes especially after all the merges pre-release. I just wanted to be sure the team is ok with the path taken in order to allocate more time.

config/network.conf.json
fboucquez: The number would be used as Uint64 when creating a LockFundsTransaction. If the property is a number it may overflow.

fboucquez:
> Listeners do not seem to work
> 
> There are no notifications for new (un)confirmed transactions

Listeners seem working, I'm announcing transactions

decentraliser:
do the snackbars show when tx get unconfirmed and confirmed?
and are the new transactions added to the transaction list?
It wasn't working when I tried

fboucquez:
Hi @decentraliser, please have another look at this PR when you have the chance

1. Fixed language selector
2. Fixed Mosaic and Namespace tables link 
3. Simplified MosaicAmountDisplay. Clients don't need to do the absolute/relative calculation or send the right divisibility. It's handled in the component. (Something similar with the ticker)  
4. Created Transaction module to handle confirmed/unconfirmed/partial transactions and blocks. I think the loaders has been simplified. Transaction dashboard loading is fine now.
5. I've revisited the listeners and they seem to be working. Maybe I've missed something? 
6. I haven't got time to test account reimport.
7. I couldn't reproduce the loading crash. What browser are you using?

decentraliser:
> 7. I couldn't reproduce the loading crash. What browser are you using?
Chrome, it seems OK now, might have been caused by one of the bugs you just fixed

decentraliser:
When creating an account, still can't see the wallet in the list, and the wallet selector does not show
image
image

decentraliser:
Others seems good. Maybe make one cycle on wallet creation / import, and then I will do an exhaustive review of all features

decentraliser:
In the dashboard, when switching wallet from the wallet selector, no_EF2D69FE94D39FE7CEC32F0C10F705F4FD400AFE931735DCF69CD35FA7C7E2FC_transactions 
shows in the transaction list
image

decentraliser:
Error is thrown when going on the transfer transaction form

Wallet.ts?40c3:266 Uncaught (in promise) Error: Public Key must be provided when calling wallet/SET_CURRENT_SIGNER!
    at _callee7$ (Wallet.ts?40c3:266)
    at tryCatch (runtime.js?98b8:45)
    at Generator.invoke [as _invoke] (runtime.js?98b8:271)
    at Generator.prototype.<computed> [as next] (runtime.js?98b8:97)
    at eval (tslib.es6.js?9ab4:74)
    at new Promise (<anonymous>)
    at __awaiter (tslib.es6.js?9ab4:70)
    at Store.SET_CURRENT_SIGNER (Wallet.ts?40c3:264)
    at Array.wrappedActionHandler (vuex.esm.js?2f62:747)
    at Store.dispatch (vuex.esm.js?2f62:438)

decentraliser:
In the dashboard balance display, clicking on the copy address icon throws

TypeError: Cannot read property 'address' of undefined

image

decentraliser:
When sending multiple mosaics, mosaics without name do not have their MosaicId prepended to their amounts
image

decentraliser:
The pencil in the mosaic table is off and there is a strange effect when clicking it

image

decentraliser:
One button icon is not showing in the peer selector
image

decentraliser:
Peer selector max size, it seems to scale following the longest string,  

It would be better to trim node names when length > some threshold
image

decentraliser:
Peer selector:
When opening the peer selector, it should automatically scroll the list to the currently selected node

decentraliser:
Changing peer: It does not seem there is a timeout system, my app stayed stuck forever in this state
image

It actually crashed my browser as well :/

fboucquez:
Hi @decentraliser, than you so much for the feedback
I've fixed some of the issues:
1) Balance copy to clipboard fixed
2) Custom mosaic ticker when no name
3) Improved how much is reloaded when peer is changed (we are not there yet for network type change). 

I've created issues from bugs reported here that are in master. I'm happy to fix them but in the next PR so it doesn't fork too much.

Other issues
1)  Error: Public Key must be provided when calling I've seen it happening sometimes but I couldn't reprodocude it.
2) Transaction table funny no transactions message. It should have been fixed. Can you reproduce?

src/components/PeerSelector/PeerSelectorTs.ts
```diff
@@ -119,29 +104,38 @@ export class PeerSelectorTs extends Vue {
...
+  get peersList(): NodeModel[] {
+    const nodeModels = this.knowNodes.filter(p => {
+      if (!this.formItems.filter) {
+        return true
+      } else {
+        return p.url.includes(this.formItems.filter)
+          || (p.friendlyName && p.friendlyName.includes(this.formItems.filter))
+      }
+    })
+    return _.sortBy(nodeModels, (a) => a.isDefault !== true, (a) => a.url)
```
decentraliser, line 126: if you really want to use lodash, you would have to add it as a dependency and import it in the file

decentraliser:
To reproduce the navigator crashing, you can try to select the node jp12.nemesis.land
this is an https node, and queried with http://
I think it would be easier to stick to the official nodes first to reduce the difficulty of this PR

src/components/PeerSelector/PeerSelectorTs.ts
fboucquez: I added the dependency before, just missed the import in this file. It fixed

fboucquez:
Hi @decentraliser @evias 

The fixes in the last commits:
1) Improved performance on peer change, exception handling and the fallback
2) Added address alias resolution to transaction list and detail
3) Improved types around transaction views
4) Mosaic Table show owned mosaic (including when the balance is zero) and other mosaics when the current user has balance.

Could you give it another go?

src/components/PeerSelector/PeerSelector.vue
```diff
@@ -98,5 +98,11 @@ export default class PeerSelector extends PeerSelectorTs {}
   .ivu-poptip-body-content {
     overflow: hidden;
   }
+  .ivu-poptip-title {
```
vincent-lee90, line 101: hey,teamate!
i will suggest that if you want to change the iview styles.please add use add `scoped`,like this `<style lang="less" scoped></style>`

src/components/ActionDisplay/ActionDisplay.vue
```diff
@@ -1,10 +1,14 @@
...
-    <span v-if="transaction.type === transactionType.TRANSFER">-></span>
-    <span>{{ descriptor }}</span>
-
-    <span v-if="needsCosignature" class="click-to-cosign">({{ $t('Click_to_cosign') }})</span>
-  </span>
+  <div>
+    <span class="col2-item overflow_ellipsis"><AddressDisplay :address="transaction.signer.address" /></span>
+    <span class="col2-item bottom overflow_ellipsis">
+      <span v-if="transaction.type === transactionType.TRANSFER"> ->
+        <AddressDisplay :address="transaction && transaction.recipientAddress" />
```
vincent-lee90, line 6: I didn't understand the type of prop `address`  is ` Address | NamespaceId`, but   the value of `transaction && transaction.recipientAddress`  is `Boolean`

src/components/AddressDisplay/AddressDisplayTs.ts
```diff
@@ -1,31 +1,62 @@
...
   }
-/// end-region computed properties getter/setter
+
+
+  /**
+   * Load transaction details
+   * @return {Promise<void>}
+   */
+  protected async loadDetails(): Promise<void> {
+    this.descriptor = ''
```
vincent-lee90, line 48: why you did  re-defined `this.descriptor = ''`?

src/components/AddressDisplay/AddressDisplayTs.ts
```diff
@@ -1,31 +1,62 @@
...
+    if (!this.address) {
+      return
+    }
+    // in case of normal transfer, display pretty address
+    if (this.address instanceof Address) {
+      this.descriptor = this.address.pretty()
+    } else if (this.address instanceof NamespaceId) {
+      this.descriptor = this.address.toHex()
+      this.descriptor = await this.$store.dispatch('namespace/RESOLVE_NAME', this.address)
```
vincent-lee90, line 57: Is it necessary to define it twice?

src/components/PeerSelector/PeerSelector.vue
```diff
@@ -4,26 +4,38 @@
...
       <div slot="title" class="title">
-        {{ $t('current_endpoint') }}：{{ currentPeer.url }}
+        {{ $t('current_endpoint') }}：{{ currentPeerInfo.url }} - {{ currentPeerInfo.friendlyName }}
       </div>
       <div slot="content">
+        <div class="inputs-container select-container">
+          <form class="input_point point_item node-inputs-container">
+            <input
+              v-model="formItems.filter"
+              :data-vv-as="$t('filter')"
```
vincent-lee90, line 14: is it `data-vv-as` or `data-vv-name` useful?

src/components/TransactionList/TransactionListTs.ts
```diff
@@ -207,21 +180,18 @@ export class TransactionListTs extends Vue {
...
-    }
+  public getCurrentTabTransactions(group: TransactionGroup): Transaction[] {
+    if (group === TransactionGroup.confirmed) return this.confirmedTransactions
+    if (group === TransactionGroup.unconfirmed) return this.unconfirmedTransactions
+    if (group === TransactionGroup.partial) return this.partialTransactions
+    if (group === TransactionGroup.all) {return [ ...this.confirmedTransactions,
+      ...this.partialTransactions, ...this.unconfirmedTransactions ]}
+
+    return []
   }
```
vincent-lee90, line 195: maybe  the `switch/case` is better，my personal point.

src/components/TransactionListOptions/TransactionListOptions.vue
```diff
@@ -0,0 +1,79 @@
...
+      color: @secondary;
+      text-decoration: underline;
+    }
+  }
+
+  .refresh-button {
+    color: @secondary;
+  }
+}
+</style>
```
vincent-lee90, line 79: I remember that i have deleted this folder TransactionListOptions

src/components/ActionDisplay/ActionDisplay.vue
evias: ACK

src/components/PeerSelector/PeerSelector.vue
evias: ACK

src/components/TransactionList/TransactionListTs.ts
evias: this code does not respect linting.

PR #219, opened 2020-04-09
the private key import times limit on the same profile(fix #214)

### LINKED ISSUE #214: The same private key can be imported into the wallet multiple times
The same private key should be limited to one import on the same profile

David0817:
fix #214


src/views/forms/FormSubWalletCreation/FormSubWalletCreationTs.ts
```diff
@@ -258,6 +265,36 @@ export class FormSubWalletCreationTs extends Vue {
...
+
+    return [...knownWallets].map(
+      ({identifier, values}) => ({
+        identifier,
+        address: values.get('address'),
+        name: values.get('name'),
+      }),
+    )
+  }
+
```
evias, line 297: Please do not copy/paste this method. If it is needed multiple times, try to relocate it appropriately.

David0817: Common data has been stored in vuex

PR #220, opened 2020-04-10
set default profile as recently logged in (fix#218)

### LINKED ISSUE #218: The default profile in login page shouldn't be the first item always
For example, I have 3 profiles  A,B,C by the order.
The default  profile always is A.
I prefer that the default profile will be B if I used profile B to login in recently.
similarly, the default profile will be C if I used profile C to login in recently.

David0817:
fix #218


evias:
Profile list should be chronologically order as they appear on database.

There is no feature request for adding a "default profile". 

Also, please do not store data in local storage unless *mapped correctly with entities*.

PR #234, opened 2020-04-13
the private key will be destroyed when the current wallet has changed

### LINKED ISSUE #222: ProtectedPrivateKeyDisplay should be destroyed on wallet change
To reproduce:

1. Reveal a private key in `Accounts` page
2. Change Wallet before the 10 seconds countdown is done
3. The private key is still revealed but should be hidden.

David0817:
fix #222

### DIFF
--- a/src/components/ProtectedPrivateKeyDisplay/ProtectedPrivateKeyDisplayTs.ts
+++ b/src/components/ProtectedPrivateKeyDisplay/ProtectedPrivateKeyDisplayTs.ts
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-import {Component, Prop, Vue} from 'vue-property-decorator'
+import {Component, Prop, Vue, Watch} from 'vue-property-decorator'
 import {Account} from 'symbol-sdk'
 // internal dependencies
 import {WalletModel} from '@/core/database/entities/WalletModel'
@@ -21,7 +21,8 @@ import {UIHelpers} from '@/core/utils/UIHelpers'
 // child components
 // @ts-ignore
 import ModalFormAccountUnlock from '@/views/modals/ModalFormAccountUnlock/ModalFormAccountUnlock.vue'
-
+const defaultCount: number = 10
+const defaultTimerDuration: number = 1000
 @Component({
   components: {
     ModalFormAccountUnlock,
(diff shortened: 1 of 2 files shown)


vincent-lee90:
the destroyed will not be called when the wallet change.
because it's the update of models instead of re-render component.
that adding a destroyed method to destroy the intervals is right.

evias:
Address should be on top of public key

David0817:
> Address should be on top of public key + addresses should be in pretty format.
This problem was fixed in PR (#255 ）

vincent-lee90:
rebased pr#177

src/components/ProtectedPrivateKeyDisplay/ProtectedPrivateKeyDisplayTs.ts
```diff
@@ -90,35 +90,25 @@ export class ProtectedPrivateKeyDisplayTs extends Vue {
   }
   /// end-region computed properties getter/setter
 
+  public init(){
```
evias, line 93: rename to `reset()`

src/components/ProtectedPrivateKeyDisplay/ProtectedPrivateKeyDisplayTs.ts
```diff
@@ -138,4 +128,15 @@ export class ProtectedPrivateKeyDisplayTs extends Vue {
...
+   * It is necessary to observe the CurrentWallet, 
+   * so that we can access to init. 
+   */
+  @Watch('wallet')
+  observeCurrentWallet(){
+    this.init()
+  }
+  public destroyed(){
+    this.init()
+  }
```
evias, line 141: semantically, this doesn't make sense. Please rename `init` to `reset()`

src/components/ProtectedPrivateKeyDisplay/ProtectedPrivateKeyDisplayTs.ts
```diff
@@ -90,35 +90,25 @@ export class ProtectedPrivateKeyDisplayTs extends Vue {
   }
   /// end-region computed properties getter/setter
 
+  public init(){
+    this.hasPlainPrivateKey = false
+    this.secondsCounter = defaultCount
+    this.countInterval && clearInterval(this.countInterval)
+    this.countInterval = null
```
evias, line 97: Please, also update `plainInformation` property to a null'd value.

vincent-lee90:
changed as the comments above.

PR #237, opened 2020-04-14
add a toString method to the instances of LibWordArray (fix #235)

### LINKED ISSUE #235: cryptoJS latest introduces breaking change with iv type
The package 'crypto-js' has updated a file 13 days ago
image
image
[https://github.com/DefinitelyTyped/DefinitelyTyped/commit/1cc65a1bdee4e5079e3c77e6ad28e6ee2c0411c9]
it will cause some errors when we reinstall the `node modules`,as the follow:
image

David0817:
fix #235


decentraliser:
this is not needed, just import crypto-js using "require"

The types are broken, but the iv object has not changed

PR #272, opened 2020-04-22
avoid the "duplicated promise" error which comes from the vue-router

### LINKED ISSUE #252: Error in log when clicking the current menu on the left menu
vue-router.esm.js?8c4f:2089 Uncaught (in promise) NavigationDuplicated {_name: "NavigationDuplicated", name: "NavigationDuplicated", message: "Navigating to current location ("/wallets/details") is not allowed", stack: "Error↵    at new NavigationDuplicated (webpack-int…node_modules/vue/dist/vue.runtime.esm.js:2178:14)"}

David0817:
( fix #252 )

### DIFF
--- a/src/components/PageNavigator/PageNavigatorTs.ts
+++ b/src/components/PageNavigator/PageNavigatorTs.ts
@@ -17,6 +17,7 @@ import {Component, Vue} from 'vue-property-decorator'
 import {mapGetters} from 'vuex'
 // internal dependencies
 import {AccountModel} from '@/core/database/entities/AccountModel'
+import { Route } from 'vue-router'
 
 @Component({
   computed: {
@@ -40,5 +41,10 @@ export class PageNavigatorTs extends Vue {
   public async logout() {
     await this.$store.dispatch('account/LOG_OUT')
     this.$router.push({name: 'accounts.login'})
+  }
+  
+  public onPageNavigate(route: Route) {
+    const isDuplicatedRoute = this.$route.matched.map(({ path }) => path).includes(route.path)
+    !isDuplicatedRoute && this.currentAccount && this.$router.push({ name: route.name }).catch(() => {/**/ })
   }
 }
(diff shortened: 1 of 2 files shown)


decentraliser:
isn't it a "NavigationDuplicated" error, as opposed to "duplicated promise"?
try adding a catch block to router.push instead
[reference](https://github.com/vuejs/vue-router/issues/2872)

image

vincent-lee90:
@decentraliser 
`try...catch...` will not work, becase the root reason is that the promise API.
this will help you [https://github.com/vuejs/vue-router/issues/2881]

decentraliser:
well, this issue is a regression, so I'm pretty sure it will work

please give it a shot

` @click="!currentAccount ? '' : $router.push({name: route.name}).catch(err => {})"`

vincent-lee90:
before i thought that's how you let me is that :
```javascript
try{
     this.currentAccount && this.$router.push({ name: route.name })
}catch(e){
}
```
and now 
 `this.currentAccount && this.$router.push({ name: route.name }).catch(err => {})"` will work.

vincent-lee90:
rebased pr177

src/components/PageNavigator/PageNavigatorTs.ts
```diff
@@ -41,4 +42,8 @@ export class PageNavigatorTs extends Vue {
     await this.$store.dispatch('account/LOG_OUT')
     this.$router.push({name: 'accounts.login'})
   }
+  
+  public onPageNavigate(route: Route) {
+    this.currentAccount && this.$router.push({ name: route.name }).catch(()=>{/**/})
```
evias, line 47: This error should not be obfuscated. If it happens, it means that likely the **button** is not being disabled when the page is active.

Please, do not obfuscate the error, make sure it is _logged_, or that _it catches only this error._

An example exploit here would be to inject `hello-world` inside `route.name`. The router will throw an unknown route exception but your code will obfuscate it.

vincent-lee90:
changed

Issue #261, opened 2020-04-18
Transaction Details Popup amount colors


fboucquez:
I'm seeing some coloring problems in the Transactions Details shown amounts: 

![image](https://user-images.githubusercontent.com/5390558/79642750-c2592400-8175-11ea-928f-096241785c55.png)

Should lost currencies be red? 

They are shown red in the transaction table

![image](https://user-images.githubusercontent.com/5390558/79642785-e7e62d80-8175-11ea-9b82-9eca7a0ed4f1.png)

What about zero values? Should they be gray/neutral?

![image](https://user-images.githubusercontent.com/5390558/79642836-2bd93280-8176-11ea-8de6-abdb1480bc55.png)

The same 0 value may be shown in red or green.

My PR fixed the Paid Fee red coloring but I want to confirm. I can fix the mosaic too.

evias:
Yes red for **outgoing**.

fboucquez:
What about 0 values? red/green/gray?

PR #287, opened 2020-04-23
Improved signer selector in transaction table

### LINKED ISSUE #274: AddressDisplay address property value error
2. path in `src/components/ActionDisplay/ActionDisplay.vue`
the type of prop `address`  is ` Address | NamespaceId`, but   the value of `transaction && transaction.recipientAddress`  is `Boolean`

fboucquez:
Fixed dashboard balance when changing the selector and the multisign account is not in the profile
User feedback improvements, values are shown quickly in the UI when swapping signers.
Outgoing mosaic values are red (gray when 0 to be defined) #261 
Fixed #274 

Same PR as https://github.com/nemfoundation/symbol-desktop-wallet/pull/280, just fixed the github history.


fboucquez:
@evias should we merge?

PR #289, opened 2020-04-24
resolve network currencies from REST network config, fixes #278

### LINKED ISSUE #278: Use rest's network configuration to resolve the network currency
At the moment, the wallet is revisiting the block/1 transactions to get information about the network currency.

https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/services/MosaicService.ts#L179

This is cumbersome and could have problems when there are 2 network currencies (like main and harvest).

Rest has a `currencyMosaicId` property here:

http://api-01.us-west-1.symboldev.network:3000/network/properties

The wallet can just load the mosaic info from that id. Note that the network properties are already loaded here https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/services/NetworkService.ts#L57
stored in the `NetworkModel`. The NetworkCurrency resolution should use the already loaded properties there.

decentraliser:


src/services/MosaicService.ts
```diff
@@ -68,14 +71,20 @@ export class MosaicService {
...
    */
   private readonly networkCurrencyStorage = new SimpleObjectStorage<NetworkCurrencyModel[]>(
     'networkCurrencyStorage')
 
+  /**
+   * Gets currency mosaicId and harvesting mosaicId from the stored network configuration
+   * @returns {MosaicId[]}
+   */
+  private getNetworkCurrenciesIds(): MosaicId[] {
+    const {networkConfiguration} = new SimpleObjectStorage<NetworkModel>('network').get()
```
fboucquez, line 84: My plan is to keep to only one place who can own the database/storage. NetworkService in this case. The service should take the network configuration as a parameter. Please provide the network configuration, the store can get that using the rootGetter.

src/services/MosaicService.ts
```diff
@@ -215,68 +229,37 @@ export class MosaicService {
...
-    if (!namespaceName) {
-      return undefined
-    }
+  private getNetworkCurrency(
+    mosaicInfo: MosaicInfo,
+    mosaicName: MosaicNames,
+  ): NetworkCurrencyModel | undefined {
+    const mosaicId = mosaicInfo.id
+
+    const namespaceName = this.getName([mosaicName], mosaicId)
```
fboucquez, line 243: is this the namespace full name? Are the parent namespace names loaded?

src/views/pages/wallets/WalletDetailsPage/WalletDetailsPage.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <div class="wallet-detail-outer-container" v-show="currentWallet">
+  <div v-show="currentWallet" class="wallet-detail-outer-container">
```
fboucquez, line 2: for some reason, this line is swapping back and forward?  is this eslint fix?

fboucquez:
Looking good, one small change and one question

src/services/MosaicService.ts
fboucquez: You can see the dependencies better this way, the network store may have not been executed yet so maybe the table is empty or not up date

src/views/pages/wallets/WalletDetailsPage/WalletDetailsPage.vue
decentraliser: I don't know why it is like that, it appears in David's PRs as well

decentraliser: just checked, it is a linting change from eslint
since v-show should be before class, I'll let it

src/services/MosaicService.ts
decentraliser: OK, Ill pass `getNetworkCurrenciesIds` as an argument provided by network store getters

I'll also add a node somewhere to say network store must be initialized before mosaic store

src/services/MosaicService.ts
```diff
@@ -176,29 +185,34 @@ export class MosaicService {
    * @param repositoryFactory tge repository factory used to load the block 1 transactions
    * @return the list of {@link NetworkCurrencyModel} found in block 1.
    */
+  // TODO move this to a service in the SDK.
   public getNetworkCurrencies(repositoryFactory: RepositoryFactory): Observable<NetworkCurrencyModel[]> {
```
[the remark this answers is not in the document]
decentraliser: ok will do

decentraliser:
/!\ Breaks the storage

network currency storage model was changed, and there are currently no database migrations

PR #301, opened 2020-04-28
Cache by Generation Hash

fboucquez:
Hey guys, in this PR I add a new storage indirection where the data is split by generation hash/network. The idea is that the data doesn't collide when swapping networks (peer selector). 

I've also added a bit of improvement around the data shown when swapping network, I know this is not really that useful now as we are going to show a big dialog saying you need to login but at least it should not show partial information from different networks in the background.

I've also updated the name of the tables, prefixed with "Cache" when they are caches.

Have a look and let me know

### DIFF
--- a/__tests__/database/backends/NetworkBasedObjectStorage.spec.ts
+++ b/__tests__/database/backends/NetworkBasedObjectStorage.spec.ts
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2020 NEM Foundation (https://nem.io)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and limitations under the License.
+ *
+ */
+import {NetworkBasedObjectStorage} from '@/core/database/backends/NetworkBasedObjectStorage'
+
+const getStorage = () => new NetworkBasedObjectStorage<number>('SomeStorageKey')
+
+describe('database/NetworkBasedObjectStorage.spec ==>', () => {
+  describe('constructor() should', () => {
+    test('create instance given no data', () => {
+      const storage = getStorage()
+      expect(storage).toBeDefined()
+    })
+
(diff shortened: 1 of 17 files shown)


decentraliser:
There is an issue with the mosaic balances

1. Log in with profile A with compatible network 1
2. Switch node to network 2 node
3. Log out, log in with profile B that is compatible with network 2
=> Balance does not show and there are 2 entries for symbol.xym in the mosaic balance list

decentraliser:
Also, sometimes, the transactions are not showing

WalletSelectionTs.ts?4dba:220 Uncaught (in promise) TypeError: Cannot read property 'address' of null
    at eval (WalletSelectionTs.ts?4dba:220)
    at Array.reduce (<anonymous>)
    at WalletSelection.mapBalanceByAddress (WalletSelectionTs.ts?4dba:220)
    at WalletSelection._callee2$ (WalletSelectionTs.ts?4dba:199)
    at tryCatch (runtime.js?98b8:45)
    at Generator.invoke [as _invoke] (runtime.js?98b8:271)
    at Generator.prototype.<computed> [as next] (runtime.js?98b8:97)
    at fulfilled (tslib.es6.js?9ab4:71)

fboucquez:
> There is an issue with the mosaic balances
> 
> 1. Log in with profile A with compatible network 1
> 2. Switch node to network 2 node
> 3. Log out, log in with profile B that is compatible with network 2
>    => Balance does not show and there are 2 entries for symbol.xym in the mosaic balance list

Fixed, network currency model needs to be reloaded when the network changes

fboucquez:
> WalletSelectionTs.ts?4dba:220 Uncaught (in promise) TypeError: Cannot read property 'address' of null
> at eval (WalletSelectionTs.ts?4dba:220)
> at Array.reduce ()
> at WalletSelection.mapBalanceByAddress (WalletSelectionTs.ts?4dba:220)
> at WalletSelection._callee2$ (WalletSelectionTs.ts?4dba:199)
> at tryCatch (runtime.js?98b8:45)
> at Generator.invoke [as _invoke] (runtime.js?98b8:271)
> at Generator.prototype. [as next] (runtime.js?98b8:97)
> at fulfilled (tslib.es6.js?9ab4:71)

WalletSelectionTs is a component used in the registration process. It loads the accounts independently from the store. We need to revisit this part.

https://github.com/nemfoundation/symbol-desktop-wallet/blob/cache_generation_hash/src/views/pages/accounts/import-account/wallet-selection/WalletSelectionTs.ts#L206

I think it's unrelated to the cache problem this PR fixes

evias:
If this breaks the account creation process, we shouldn't merge it for 0.9.9 release because everyone is going over the account creation process after install.

decentraliser:
did you find a big relative to account creation process @evias?

PR #294, opened 2020-04-27
SDK and testnet upgrade

decentraliser:
* upgrade SDK to v0.18.0, fixes #276
* remove AESEncryptionService, fixes #277
* Testnet upgrade, fixes #292
* Move user data storage folder to `/home/.symbol-desktop-wallet`
* Rename accounts to profiles, wallets to accounts #304

PR #308, opened 2020-04-29
Storage Versioning

fboucquez:
Hi guys, I've implemented versioning.

The idea is that each table at top level has a version. A version per table, not per row like it used to be. Like this:

image

A table/storage may need more migrations that other tables. The migrations are handle per table/storage.

Then when the Storage object is created, the migration functions can be applied (if the stored version is old). Migration function is a function from A to A'. We are going to maintain a list of Migration functions when required. (Note that A can be a single value, a list, a complex object, etc)

Storage clients (like services) don't know about versioning, they just load/save/delete the model objects (not the top-level wrapper). 

Note that per model storage/table are singleton, they are created once just to run the migration once (if needed) per execution

Implements https://github.com/nemfoundation/symbol-desktop-wallet/issues/305

### DIFF
--- a/__tests__/database/backends/NetworkBasedObjectStorage.spec.ts
+++ b/__tests__/database/backends/NetworkBasedObjectStorage.spec.ts
@@ -14,8 +14,9 @@
  *
  */
 import {NetworkBasedObjectStorage} from '@/core/database/backends/NetworkBasedObjectStorage'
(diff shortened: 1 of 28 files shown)


decentraliser:
I'm not going to merge it, and gonna copy it "manually"
I'll do a dedicated commit so it is isolated

decentraliser:
integrated manually in SDK and testnet upgrade #294

PR #337, opened 2020-05-08
AggregateBoundedAnnounceService

decentraliser:


src/services/AggregateBoundedAnnounceService.ts
```diff
@@ -0,0 +1,154 @@
...
+
+// internal dependencies
+import { AbstractService } from './AbstractService'
+import { BroadcastResult } from '@/core/transactions/BroadcastResult'
+
+// configuration
+import conf from '@/../config/app.conf.json'
+const { HASH_LOCK_CONFIRMATION_TIMEOUT } = conf.constants
+
+export class AggregateBoundedAnnounceService extends AbstractService {
```
fboucquez, line 29: I would stop using AbstractService. It doesn't provide anything helpful

PR #354, opened 2020-05-14
changed url validator

### LINKED ISSUE #291: Enter a crafted URL in Settings -> Explorer URL cause the Desktop Wallet to freeze.
Hey Guys, 

I come from your bug bounty program: https://hackerone.com/symbol.  I am not sure if my finding is in scope, I don't wanna risk my report get close as Informative or N/A and the finding doesn't really contain any sensitive information, so I decided to post here.

#### Why I think this is in scope: The finding falls into the low category:
Low
These would be mostly user experience level issues that render a end user to be stuck or applications to be useless. Typically this could be something like a user submitting to a form that breaks and user gets locked into a certain workflow until killing/quitting an application

#### Why I think this might not in scope. 
Out of scope features: Offline settings

Please let me know if this is a valid finding. 

### Summary
Enter a crafted URL in `Settings -> Explorer URL` cause the Desktop Wallet to freeze.

I believe this is caused by the regular expression used to validate the Explorer URL in `src/core/validation/validators/UrlValidator.ts`. The regular expression is so complex that it consumes lots of CPU when it runs against specific URLs.

```
  public static validate(value: any): boolean {
    const pattern = new RegExp('^(https?:\\/\\/)?' + // protocol
       '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.?)+[a-z]{2,}|' + // domain name
       '((\\d{1,3}\\.){3}\\d{1,3}))' + // ip (v4) address
       '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*' + // port
       '(\\#[-a-z\\d_]*)?$','i')
    return pattern.test(value)
  }
```

For more information about regex DOS: https://owasp.org/www-community/attacks/Regular_expression_Denial_of_Service_-_ReDoS

### Steps To Reproduce:
1. In the Setting interface, set the Explorer URL to "http://aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.com/a?a=1"
2. Click Confirm

In the payload above, the question mark `?` is the character that causes the issue. I will expect to see a user accidentally enter or paste a question mark into the URL, so this is something that would likely happen. The situation gets worse if you try it in the Dev environment that runs the app in a browser.

### Supporting Material/References:

Screen Shot 2020-04-25 at 6 50 50 PM

Screen Shot 2020-04-25 at 6 51 01 PM

vincent-lee90:
fixes #291
I hoped that our url validate rules should be simple and effective.
Complex validate rules will cause time out
Maybe there are some other rules I didn't match, If you found them, you'll be welcome to tell me .

### DIFF
--- a/__tests__/core/validation/validators/UrlValidator.spec.ts
+++ b/__tests__/core/validation/validators/UrlValidator.spec.ts
@@ -1,0 +1,59 @@
+import { UrlValidator } from '@/core/validation/validators'
+
+describe('UrlValidator.validate should', () => {
+  describe('return false when passing in', () => {
+    test('http://', () => {
+      expect(UrlValidator.validate('http')).toBeFalsy()
+    })
+    test('https://', () => {
+      expect(UrlValidator.validate('http://')).toBeFalsy()
+    })
+    test('https://foo', () => {
+      expect(UrlValidator.validate('https://q')).toBeFalsy()
+    })
+    test('https://foo..', () => {
+      expect(UrlValidator.validate('https://foo..')).toBeFalsy()
+    })
+    test('https://foo.-', () => {
+      expect(UrlValidator.validate('https://foo.-')).toBeFalsy()
+    })
+    test('https://foo--', () => {
+      expect(UrlValidator.validate('https://foo--')).toBeFalsy()
+    })
+    test('https://192.0.0', () => {
+      expect(UrlValidator.validate('https://192.0.0')).toBeFalsy()
+    })
+    test('https://goo_gle.com', () => {
+      expect(UrlValidator.validate('https://goo_gle.com')).toBeFalsy()
+    })
+    test('localhost', () => {
+      expect(UrlValidator.validate('localhost')).toBeFalsy()
+    })
+  })
+  describe('return true when passing in', () => {
+    test('google.com', () => {
+      expect(UrlValidator.validate('google.com')).toBeTruthy()
+    })
+    test('http://google.com', () => {
+      expect(UrlValidator.validate('http://google.com')).toBeTruthy()
+    })
+    test('https://google.com', () => {
+      expect(UrlValidator.validate('https://google.com')).toBeTruthy()
+    })
+    test('https://goo-gle.com', () => {
+      expect(UrlValidator.validate('https://goo-gle.com')).toBeTruthy()
+    })
+    test('https://goo.gle.com', () => {
+      expect(UrlValidator.validate('https://goo.gle.com')).toBeTruthy()
+    })
+    test('192.0.0.1:8080', () => {
+      expect(UrlValidator.validate('192.0.0.1:8080')).toBeTruthy()
+    })
+    test('localhost:8080', () => {
+      expect(UrlValidator.validate('localhost:8080')).toBeTruthy()
+    })
+    test('http://api-01.eu-west-1.0941-v1.symboldev.network:3000', () => {
+      expect(UrlValidator.validate('http://api-01.eu-west-1.0941-v1.symboldev.network:3000')).toBeTruthy()
+    })
+  })
+})
--- a/src/core/validation/validators/UrlValidator.ts
+++ b/src/core/validation/validators/UrlValidator.ts
@@ -23,15 +23,15 @@ export class UrlValidator {
    * @param {*} value
    * @returns {boolean}
    */
-  public static validate(value: any): boolean {
-    const pattern = new RegExp(
-      '^(https?:\\/\\/)?' + // protocol
-      '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.?)+[a-z]{2,}|' + // domain name
-      '((\\d{1,3}\\.){3}\\d{1,3}))' + // ip (v4) address
-      '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*' + // port
-        '(\\#[-a-z\\d_]*)?$',
-      'i',
-    )
+  public static validate(value: string): boolean {
+    const regStr =
+      '^' + // position at start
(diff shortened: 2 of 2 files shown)


src/core/validation/validators/UrlValidator.ts
```diff
@@ -25,11 +25,7 @@ export class UrlValidator {
    */
   public static validate(value: any): boolean {
     const pattern = new RegExp(
-      '^(https?:\\/\\/)?' + // protocol
-      '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.?)+[a-z]{2,}|' + // domain name
-      '((\\d{1,3}\\.){3}\\d{1,3}))' + // ip (v4) address
-      '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*' + // port
-        '(\\#[-a-z\\d_]*)?$',
```
evias, line 32: i would prefer to keep the above regular expression as it is well **documented**.

evias: I am sorry @vincent-lee90 I didnt see the Issue about this but it seems this regular expression is the cause of performance issues.

Please re-introduce your regular expression and keep the unit tests above :+1:

vincent-lee90:
first , the previous regular expression could not pass the unit test I wrote
second, I use a more precise scope to avoid backtracking problems caused by regular expression mismatches

__tests__/core/validation/validators/UrlValidator.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+      expect(UrlValidator.validate('https://foo..')).toBeFalsy()
+    })
+    test('https://foo.-', () => {
+      expect(UrlValidator.validate('https://foo.-')).toBeFalsy()
+    })
+    test('https://foo--', () => {
+      expect(UrlValidator.validate('https://foo--')).toBeFalsy()
+    })
+    test('https://192.0.0', () => {
+      expect(UrlValidator.validate('https://foo--')).toBeFalsy()
```
decentraliser, line 24: seems 'https://192.0.0' should be in validate()

__tests__/core/validation/validators/UrlValidator.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    })
+    test('http://api-02.eu-central-1.0941-v1.symboldev.network:3000', () => {
+      expect(UrlValidator.validate('http://api-02.eu-central-1.0941-v1.symboldev.network:3000')).toBeTruthy()
+    })
+    test('http://api-01.ap-southeast-1.0941-v1.symboldev.network:3000', () => {
+      expect(UrlValidator.validate('http://api-01.ap-southeast-1.0941-v1.symboldev.network:3000')).toBeTruthy()
+    })
+    test('http://api-01.us-west-1.0941-v1.symboldev.network:3000', () => {
+      expect(UrlValidator.validate('http://api-01.us-west-1.0941-v1.symboldev.network:3000')).toBeTruthy()
+    })
```
decentraliser, line 72: I think only one of these URLS should be enough

__tests__/core/validation/validators/UrlValidator.spec.ts
vincent-lee90: u r right! thanks

__tests__/core/validation/validators/UrlValidator.spec.ts
vincent-lee90: > I think only one of these URLS should be enough

I've considered using only one of them before, but in order to ensure that all our current nodes can be connected, there will be no other accidents

decentraliser: Then maybe importing nodes that are in the network configuration object would be better? Nodes URLs will bet changed many times, if this test file is not maintained with up-to-date URLs, it won't really make sense anymore in a close future

vincent-lee90: Even if one day these URLs are out of date, these out of date URL unit tests still make sense

__tests__/core/validation/validators/UrlValidator.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+    test('google.com', () => {
+      expect(UrlValidator.validate('google.com')).toBeTruthy()
+    })
+    test('http://google.com', () => {
+      expect(UrlValidator.validate('http://google.com')).toBeTruthy()
+    })
+    test('https://google.com', () => {
+      expect(UrlValidator.validate('https://google.com')).toBeTruthy()
+    })
+    test('https://goo-gle.com', () => {
```
decentraliser, line 37: test('https://goo_gle.com', () => {
      expect(UrlValidator.validate('https://goo_gle.com')).toBeTruthy()
    })

please add this test and fix the validation, seems the validator doesn't allow for underscores

__tests__/core/validation/validators/UrlValidator.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+
+describe('UrlValidator.validate should', () => {
+  describe('return false when passing in', () => {
+    test('http://', () => {
+      expect(UrlValidator.validate('http')).toBeFalsy()
+    })
+    test('https://', () => {
+      expect(UrlValidator.validate('http://')).toBeFalsy()
+    })
+    test('https://foo', () => {
```
evias, line 11: nothing wrong about this kind of urls ... ? e.g. `localhost`

__tests__/core/validation/validators/UrlValidator.spec.ts
evias: there is no need to copy all these URLs in there, please remove them and use only one.

PR #359, opened 2020-05-18
announce bonded with tx service and add timeout, fixes #52

### LINKED ISSUE #52: Aggregate bounded announce: create a timeout system
When announcing a hash lock, a timeout process should start, waiting for the hash lock confirmation

if confirmation is not received before the timeout, the confirmed listener should be closed

decentraliser:
1 point for team rxjs

### DIFF
--- a/src/store/Account.ts
+++ b/src/store/Account.ts
@@ -24,8 +24,12 @@ import {
   RepositoryFactory,
   SignedTransaction,
   Transaction,
+  TransactionService,
 } from 'symbol-sdk'
 import { of, Subscription } from 'rxjs'
+import { catchError, map, timeoutWith, tap, mapTo } from 'rxjs/operators'
+import * as _ from 'lodash'
+
 // internal dependencies
 import { $eventBus } from '../events'
 import { RESTService } from '@/services/RESTService'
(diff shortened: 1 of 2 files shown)


src/store/Account.ts
```diff
@@ -511,39 +513,30 @@ export default {
...
+      await listener.open()
+
+      return transactionService
+        .announceHashLockAggregateBonded(signedLock, signedPartial, listener)
+        .pipe(
+          catchError((error) => of(new BroadcastResult(signedPartial, false, error))),
+          mapTo(new BroadcastResult(signedPartial, true)),
+          timeoutWith(
+            10_000_000, // 10 minutes timeout
+            of(new BroadcastResult(signedPartial, false, 'Announce timed out')),
```
evias, line 532: `Aggregate transaction broadcast timed out`

src/store/Account.ts
```diff
@@ -511,39 +513,30 @@ export default {
...
+      const listener = repositoryFactory.createListener()
+      await listener.open()
+
+      return transactionService
+        .announceHashLockAggregateBonded(signedLock, signedPartial, listener)
+        .pipe(
+          catchError((error) => of(new BroadcastResult(signedPartial, false, error))),
+          mapTo(new BroadcastResult(signedPartial, true)),
+          timeoutWith(
+            10_000_000, // 10 minutes timeout
```
evias, line 531: Seems like a very long period to be inspecting blocks ; Wouldn't 2 minutes suffice? If there is no block including the hash lock after 2 minutes, there is already something wrong, that's what i mean

decentraliser:
Ok, I can put this timeout in the app config also

src/store/Account.ts
```diff
@@ -511,39 +513,30 @@ export default {
...
-        return new BroadcastResult(signedPartial, false, e.toString())
-      }
+      // - prepare REST parameters
+      const repositoryFactory = rootGetters['network/repositoryFactory'] as RepositoryFactory
+      const transactionHttp = repositoryFactory.createTransactionRepository()
+      const receiptsHttp = repositoryFactory.createReceiptRepository()
+      const transactionService = new TransactionService(transactionHttp, receiptsHttp)
+      // - prepare scoped *confirmation listener*
+      const listener = repositoryFactory.createListener()
+      await listener.open()
```
fboucquez, line 523: Why not reuse the network's listener? 
If you create a new listener, I won't mix promise and observable. Try creating observable from the wait promise and pipe it with the other observables.

BTW, I'm moving the transaction logic out of the account. I definitely add the new timeout in my new PR.

PR #364, opened 2020-05-18
delete unused translations entries

### LINKED ISSUE #311: Delete unused translations entries
Delete unused language keys in language .json files

David0817:
( fix #311 )
 * remove unused
  * specification keys name
  * order by keys ASC


decentraliser:
Could you include a script in package.json for finding duplicates entries?

David0817:
> please as mentioned by @decentraliser, add a package.json script to identify keys that are never used in `src/`
> 
> `"lint:i18n": "...."`

There is a problem when i use plug-ins for testing.
When i18n is used in a function, the variable is marked as missing and the actual value is marked as unused.
I don't have a good way to solve it,  do you have any suggestions 
@evias @decentraliser

decentraliser:
Yeah, I'm pretty sure a perfect solution doesn't exist .

As long as the script is not in the lint / build workflow, it's ok. Try to
check how other people handle this case

David0817:
Now in pr, I manually clean up the useless i18n keys, using ASC sorting and normalizing the naming of keys. ASC sorting can provide a convenient way to sort out more languages in the future, and if we follow this rule later, it will be easier to manage i18n files. However, there is no lint method that works for i18n for the time being, because i18n may appear in the enumeration of TS or the variable of VUE in the program. Even if the check can be achieved, it needs to be proofread manually, not to mention the fact that none of the three languages are complete.

therefore.
Please help review and merge PR, for further optimization on this basis. For i18n file script management, I will build a new issue for processing, otherwise more merge may invalidate the current work.

we have not found a solution, if you have a good script implementation,  please recommend, I will take the time to study the implementation and complete the modification.

PR #368, opened 2020-05-19
Recommended Fee

fboucquez:
- Transaction Command
- Removed stage/signed transactions from account
- Remove values's from map
- Simplified Trezor confirmation screen
- Simplfied transaction forms


evias:
- Network Median => `Normal`
- Network Highest => `Fastest`
- Free => `No fee`

fboucquez:
Fastest, do you want to show them? It could be pretty expensive
I can take some time to add the calculated fee on the form screen but if we are going to change the transaction workflow we can apply it then. I think we can take advantage of the general refactoring now with this PR, especially in the ledger branch

evias:
rebased and fixed TransactionService conflict

PR #375, opened 2020-05-21
Added scoped

### LINKED ISSUE #273: Add scoped less import for PeerSelector
1. path in `src/components/PeerSelector/PeerSelector.vue` ,
i will suggest that if you want to change the iview styles in this component.please add `scoped`,like this `<style lang="less" scoped></style>`

### LINKED ISSUE #307: Add scoped to the style of all components
Hey, friends.
At present, the project has reached a very good stage, but due to the previous rapid development, there are still some problems left. I hope we can improve these problems.
# About Less
With the development of desktop wallet, more and more components will appear. There may be some style conflicts between components. For example, you add `scoped` to the newly developed component styles, but you will find that the components you write are affected by the style files you don't know where, because we didn't add `scoped` to some old component styles. like [here](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/components/MosaicBalanceList/MosaicBalanceList.less#L258),[here ](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/components/PeerSelector/PeerSelector.vue#L86),[here](https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/components/LanguageSelector/LanguageSelector.vue#L13) and so on.  
the first one it changed the styles of iview tables, but the iview components we use in other components do not need these styles.what should we change at this time?  
there are many places like the second and the third, this will cause overall style pollution.  
for example, we are developing a component, but the default style is that when we finished the template
image  
We haven't started writing component styles yet.  
There are many similar problems, I find some as the follows:  
image  
image
  
About more details, you can see [here](https://vuejs.org/v2/style-guide/index.html#Component-style-scoping-essential)
We can see that Vue officials highly recommend that we do  that  add styles `scoped`,besides, common.less  should be updated, a lot of styles are out of date because of the transformation of our code, many old codes are very confusing to us.

## what i will do?
     1. add scoped to all components   and  adjust style
     2. update some common.less 

## benefits?
There won't be many strange problems in CSS, and even if they do, we can easily fix them in CSS scope, and in most cases, we don't need to consider affecting other elements,because of the `scoped`,  and we could global change global style, local change local style.

vincent-lee90:
Added scoped to components(fixed #307)(fixed #273)


decentraliser:
There are several  //unused comments, flagging CSS classes that are effectively not used.

Is there a reason for keeping them in the code?

vincent-lee90:
> There are several //unused comments, flagging CSS classes that are effectively not used.
> 
> Is there a reason for keeping them in the code?
hahaha, because this PR is the first PR in my next series
In my next PR, these `//unused` will be cleared by me

decentraliser:
There is an overflow issue on the profile name input border
image

src/components/WindowControls/WindowControls.vue
```diff
@@ -11,7 +11,8 @@
 
 <script lang="ts">
```
decentraliser, line 12: AFAIK the `WindowControls` was used to add window controls to the Windows build

But when we worked on the build, I set the Windows build window controls to be the native Windows controls

So, if this component is not necessary for the Mac build, we can delete it

vincent-lee90:
> There is an overflow issue on the profile name input border
> image
Sorry , i can't see your picture because of the network .
But don't worry, 

> There is an overflow issue on the profile name input border
> image

Don't worry. I can imagine that this problem is caused by the nonstandard CSS code. My PR will deal with it later

src/components/WindowControls/WindowControls.vue
vincent-lee90: I think you can submit an issue to track it, not in this PR, right?

src/views/layout/PageLayout/PageLayout.common.less
```diff
@@ -288,9 +256,56 @@
...
     }
   }
-
+  //unused
   .node-input-container {
     width: 100%;
     margin-right: 20px;
   }
 }
+.mac.wrap {
```
decentraliser, line 265: Can you please rename this class?

It is named `mac` because there was a `windows` class in the past

A more generic name like `app-container` would be more appropriate and avoid confusion in the future

vincent-lee90: > generic name like

Certainly.

I did this before, but @evias asked me to keep the purity of this pr. I will deal with what you said in the PR later

decentraliser:
The login input border bug seems to only occur on smaller resolutions, so I guess it's ok

PR #386, opened 2020-05-23
upgrade to 951 testnet, SDK 0.19.1, fixes #385

### LINKED ISSUE #385: Upgrade SDK to v0.19.1

decentraliser:


src/core/database/storage/NetworkModelStorage.ts
```diff
@@ -24,6 +30,15 @@ export class NetworkModelStorage extends VersionedNetworkBasedObjectStorage<Netw
...
 
   private constructor() {
-    super('networkCache')
+    super('networkCache', [
+      {
+        description: 'Update networkCache to 0.9.5.1 network',
+        migrate: (from: any) => {
+          // remove all pre-0.9.5.1 networks
+          const networks = Object.keys(from)
+          networks.map((networkGenHash) => resetNetwork(networkGenHash))
```
fboucquez, line 39: The migration is a standalone function that converts from the old value to a new value. It does not need services (storage should not depend on higher levels components like services)

In this case, because you just want to clear out the table the function is:

`migrate: (from: any) => return undefined`

undefined means the data is going to be removed

https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/core/database/backends/VersionedObjectStorage.ts#L57

src/core/database/storage/NodeModelStorage.ts
```diff
@@ -24,6 +35,14 @@ export class NodeModelStorage extends VersionedObjectStorage<NodeModel[]> {
   public static INSTANCE = new NodeModelStorage()
 
   private constructor() {
-    super('node')
+    super('node', [
+      {
+        description: 'Update node to 0.9.5.1 network',
+        migrate: (from: any) => {
+          // remove all pre-0.9.5.1 nodes
+          resetPeers(from)
```
fboucquez, line 43: Same, just return undefined.

src/core/database/storage/NetworkModelStorage.ts
evias: yes i will apply 0.19.2 before release :+1:

evias: ok thanks for the notes on migration

evias: fixed with e12ff0fa

src/core/database/storage/NodeModelStorage.ts
evias: fixed with e12ff0fa

src/services/NetworkService.ts
```diff
@@ -168,7 +168,7 @@ export class NetworkService {
     const repositoryFactory = new RepositoryFactoryHttp(url)
```
fboucquez, line 168: How the wallet creates the RepositoryFactoryHttp is going to be simplified in the next SDK. The createLister won't need to be patched

PR #433, opened 2020-06-22
Update en-US.json

44uk:
Duplicate string key `current_network` in line 137 and line 556. Removed later.


evias:
fixed with harvesting PR :+1:

44uk:
OK, I got it.

PR #434, opened 2020-06-23
add condition for displaying the prompt text in ModalTransactionCosignature

### LINKED ISSUE #410: Improve the prompt text of the detail page in the multi-signature verification process
When all co-signers of a multi-signature transaction have signed off, the prompt text on the detail page should be changed.

图片5

vincent-lee90:
fixes #410 
Now the prompt text won't  be hidden until min required signatures have been received


vincent-lee90:
> @vincent-lee90
> `Co-signatures received` list is missing. This list is important because it informs which co-signer(s) out of all co-signers participated in the approval.
> 
> After the minimum required co-signers have signed, then instead of hiding the prompt text, might be better to change it to something like "Received the minimum required co-signatories".

This one does look better!

PR #441, opened 2020-06-29
remove digital animation

### LINKED ISSUE #358: Network Static Panel showing accounts instead of transactions
The UI states **Transactions** but it is actually showing **Accounts**

image

It mean just be a label problem but do we want to show accounts, transactions, or both?

https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/language/en-US.json#L239

https://github.com/nemfoundation/symbol-desktop-wallet/blob/master/src/components/NetworkStatisticsPanel/NetworkStatisticsPanel.vue#L6

Would it be possible to remove Animation and use a static number? I think it is a bit annoying to wait for the counter to finish every time you go back to the dashboard. If it's a loading issue, to avoid jumping from 0 to a real number, we could add a loading spinner, storage cache and/or prefetch data while logging in.

David0817:
fix #358

### DIFF
--- a/src/components/NetworkStatisticsPanel/NetworkStatisticsPanel.vue
+++ b/src/components/NetworkStatisticsPanel/NetworkStatisticsPanel.vue
@@ -33,7 +33,7 @@
         </div>
         <img src="@/views/resources/img/monitor/peer_number.png" />
         <span class="txt_info">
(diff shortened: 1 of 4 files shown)


evias:
rebased and fixed in #462. 

Thanks @David0817 for participating :+1:

PR #446, opened 2020-06-30
Dev ledger with version 0.10.1

thanhhang-fds:
Symbol-desktop-wallet is integrated with Ledger


evias:
blocked after Ledger review of NIS integration, needs rework.

PR #450, opened 2020-07-06
Add the exported data to the CSV file

### LINKED ISSUE #416: csv-export for the desktop-wallet
The desktop-wallet should contain a csv export for transactions and harvesting. This also has something to do with usability.
We should give the ordinary user the possibility to retrieve and store all transactions of his account without much effort and without being dependent on anyone else.

David0817:
fix #416


fullcircle23:
Also, the style, microcopy, and position of the `Download` button need to be revisited separately. This applies to the Harvesting `Show wizard` button and potentially other components as well to achieve a clean, consistent and aesthetically pleasing design.

Issue #459, opened 2020-07-16
Wrong link to explorer for account.


cryptoBeliever:
Double slash is added to explorer link.
For example:

http://explorer-96x.symboldev.network//account/TAUTUPN57JYRZGVQ26IMX4RJ7EXU4CMLZWNGKEA

It causes blank page on explorer side.

PR #467, opened 2020-07-16
fix explorer link #469

bassemmagdy:
fixing #459 and #469

### DIFF
--- a/src/components/AccountLinks/AccountLinksTs.ts
+++ b/src/components/AccountLinks/AccountLinksTs.ts
@@ -39,7 +39,7 @@ export class AccountLinksTs extends Vue {
   /// end-region computed properties getter/setter
 
   public get explorerUrl() {
-    return this.explorerBaseUrl.replace(/\/+$/,'')  + '/account/' + this.account.address
+    return this.explorerBaseUrl.replace(/\/+$/, '') + '/account/' + this.account.address
   }
 
   public get faucetUrl() {
--- a/src/components/AccountLinks/AccountLinksTs.ts
+++ b/src/components/AccountLinks/AccountLinksTs.ts
@@ -39,7 +39,7 @@ export class AccountLinksTs extends Vue {
   /// end-region computed properties getter/setter
 
   public get explorerUrl() {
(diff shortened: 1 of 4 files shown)


src/components/AccountLinks/AccountLinksTs.ts
```diff
@@ -39,7 +39,7 @@ export class AccountLinksTs extends Vue {
   /// end-region computed properties getter/setter
 
   public get explorerUrl() {
-    return this.explorerBaseUrl + '/account/' + this.account.address
+    return this.explorerBaseUrl + 'account/' + this.account.address
```
evias, line 42: this implies that the configured explorerBaseUrl comes with an ending `/`.

Please use something like: `return this.explorerBaseUrl.replace(/\/+$/,'') + '/account';`

as the above will permit configuration values to hold with or without slash.

src/components/TransactionDetails/TransactionDetailRow/TransactionDetailRow.vue
```diff
@@ -5,9 +5,7 @@
     </div>
     <div class="transaction-details-row-value-container">
       <span v-if="label === 'hash' || label === 'inner_transaction_hash'">
-        <a class="url_text" target="_blank" :href="(explorerBaseUrl + '/transaction/' + item.value)">{{
-          item.value
-        }}</a>
+        <a class="url_text" target="_blank" :href="(explorerBaseUrl + 'transaction/' + item.value)">{{ item.value }}</a>
```
evias, line 8: same comment, probably good to have a component level method do this job, though :+1:

src/components/TransactionList/TransactionRow/TransactionRow.vue
```diff
@@ -33,7 +33,7 @@
         <a
           class="url_text"
           target="_blank"
-          :href="(explorerBaseUrl + '/transaction/' + transaction.transactionInfo.hash)"
+          :href="(explorerBaseUrl + 'transaction/' + transaction.transactionInfo.hash)"
```
evias, line 36: same here, component level method

PR #481, opened 2020-07-27
Multisig min approval removal fix

### LINKED ISSUE #372: Failure multisig min setting out of range
After adjusting the number of cosigners needed to sign a transaction:
Min approval from 2 to 0
and min removal from 3 to 0

Should bring up a message that adjusting to 0 is not possible if there are still cosigners?

image

Same problem when removing all cosigners but not setting approval and removal to 0.

The transaction is successful, but after signing by the cosigner it gives the error.

bassemmagdy:
fixes #372


bassemmagdy:
@evias rebased

Issue #357, opened 2020-05-14
The creation fee for creating mosaic should be shown to the accounts


Alicely:
To create a mosaic, accounts have to pay a fee in addition to the transaction fee.The fees will be deducted from the account’s balance after the announcement of a valid MosaicDefinitionTransaction.However, the creation fee is not shown in the creation process and transaction details.（relate#356）

cryptoBeliever:
Estimated rental fee is presented.

PR #484, opened 2020-08-03
Display current rental fee

### LINKED ISSUE #356: The rental fee for creating namespace or extending its duration should be shown to the accounts
To create a namespace or to extend its duration, accounts have to pay a rental fee in addition to the transaction fee. Both fees will be deducted from the account’s balance after the announcement of a valid NamespaceRegistrationTransaction.However, the rental fee is not shown in the creation process and transaction details.When this rental fee is deducted, users will be confused.

bassemmagdy:
should resolve #356 #357

### DIFF
--- a/src/components/RentalFees/RentalFeeTs.ts
+++ b/src/components/RentalFees/RentalFeeTs.ts
@@ -1,0 +1,35 @@
+import { Component, Vue, Prop, Watch } from 'vue-property-decorator'
+//@ts-ignore
+import MosaicAmountDisplay from '@/components/MosaicAmountDisplay/MosaicAmountDisplay.vue'
+//@ts-ignore
+import FormRow from '@/components/FormRow/FormRow.vue'
+
+export type RentalFeesType = 'mosaic' | 'root-namespace' | 'child-namespace'
+@Component({
+  components: {
+    MosaicAmountDisplay,
+    FormRow,
+  },
+})
+export class RentalFeeTs extends Vue {
+  @Prop({ required: true }) rentalType: RentalFeesType
+  @Prop() duration: number
+  @Prop({ required: true }) rentalFee: number
+
+  /**
+   * @description: return effectiveFee according to prop 'rentalType'
+   */
+  get rentalFeeAmount(): number {
+    let feeAmountRaw: number = this.rentalFee
+    return this.rentalType == 'root-namespace' ? (feeAmountRaw *= this.duration) : feeAmountRaw
+  }
+
+  @Watch('rentalFeeAmount')
+  onRentalFeeAmountChange() {
+    this.$store.commit('network/currentRentalFee', this.rentalFeeAmount)
+  }
+
+  destroyed() {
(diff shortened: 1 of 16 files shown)


src/components/RentalFees/RentalFeeTs.ts
```diff
@@ -0,0 +1,35 @@
...
+import FormRow from '@/components/FormRow/FormRow.vue'
+
+export type RentalFeesType = 'mosaic' | 'root-namespace' | 'child-namespace'
+@Component({
+  components: {
+    MosaicAmountDisplay,
+    FormRow,
+  },
+})
+export class RentalFeeTs extends Vue {
```
evias, line 14: recommendation to use store getters instead of statically passed in `rentalFees` component-level property

src/core/transactions/TransactionView.ts
```diff
@@ -137,25 +136,39 @@ export abstract class TransactionView<T extends Transaction> {
...
+      return [
+        {
+          key: 'max_fee',
+          value: {
+            amount: this.transaction.maxFee.compact() || 0,
+            color: 'red',
+          },
+          isMosaic: true,
+        },
+        {
```
evias, line 163: does this **have** to be added for **all** transaction types ? i.e. it is being added to the generic `TransactionView`, while rental fees are available only for a small count of transaction types, would there be a way to extend this method by child classes that would indeed have rental fees attached, only?

src/store/Network.ts
```diff
@@ -67,6 +67,7 @@ interface NetworkState {
   currentHeight: number
   subscriptions: Subscription[]
   transactionFees: TransactionFees
+  currentRentalFee?: number
```
evias, line 70: recommendation to use `rentalFeeEstimations`

src/store/Network.ts
```diff
@@ -268,6 +274,12 @@ export default {
       }
     },
 
+    RENTAL_FEE({ rootGetters }): Promise<RentalFees> {
```
evias, line 277: - start all REST querying store actions with `REST_`
- recommendation to use name `REST_NETWORK_RENTAL_FEES`

src/views/forms/FormMosaicDefinitionTransaction/FormMosaicDefinitionTransaction.vue
```diff
@@ -33,7 +33,7 @@
             </div>
           </template>
         </FormRow>
-
+        <RentalFee :rental-type="'mosaic'" :rental-fee="formItems.rentalFees"></RentalFee>
```
evias, line 36: seems that `duration` is not being passed over to the component here?

bassemmagdy: does duration affect fees while creating new mosaics?  or just while creating new root namespaces?

bassemmagdy:
updates in #489

PR #516, opened 2020-09-28
Seprate wallets fixes #221

### LINKED ISSUE #221: Separate HD Wallets from Private Key wallets
Import account page:
- Add a link around "more options are being worked on" to import a private key

Account creation:
- It should be possible to create a profile with a private key

Wallets import
- When using an HD profile, it should not be possible to import a private key
- When using a private key profile, it should not be possible to import a private key nor a wallet

bassemmagdy:
Separating hd-wallets from private key ones by allowing each user to create a new profile by importing his own private key. fixes #221


bassemmagdy:
@evias please note that this pr is based on @yilmazbahadir's branch for importing mnemonic and that's why it contains those big changes.

evias (approved):

Issue #551, opened 2020-10-08
Multisig Account Page - shouldn't allow to send when there is no change


yilmazbahadir:
On the multisig account page, send button is enabled by default and it allows you to send the transaction even when nothing is changed. The send button should be visible but disabled by default and get enabled when anything changes subject to a transaction.

cryptoBeliever:
@yilmazbahadir it's fixed so I'm closing.

Issue #550, opened 2020-10-07
Multisig Account Page - multiple cosignatories removal should be disallowed


yilmazbahadir:
Wallet app shouldn't allow users to remove multiple cosignatories since only one cosignatory removal is allowed in a transaction.

![Screenshot 2020-10-07 at 09 26 02](https://user-images.githubusercontent.com/6256269/95362432-fbbd5180-08c5-11eb-82a2-d7331b08937b.png)

cryptoBeliever:
@yilmazbahadir it's fixed so I'm closing.

Issue #543, opened 2020-10-05
Explorer link is not working 


surekabpm:
Defect description : Explorer link is broken

Steps to reproduce : 
1. create an account and login to the account 
2. navigate to accounts tab
3.click the link Explorer "open in explorer" under useful links
4.it should directly open in browser (chrome), but it does'nt ,it is asking for app to open

![image](https://user-images.githubusercontent.com/49143210/95066574-338a9480-0720-11eb-861d-1a2a5e356019.png)

evias:
The problem is happening because of previously stored `settings` that hold an explorerUrl.

A settings migration could work.

Issue #540, opened 2020-10-04
Unable to convert account to multi-signature


OlegMakarenko:
![image](https://user-images.githubusercontent.com/33131259/95026802-a2260e80-069c-11eb-95bd-dd288d5bd59b.png)
![image](https://user-images.githubusercontent.com/33131259/95026846-ddc0d880-069c-11eb-9566-aa60c5982335.png)
![image](https://user-images.githubusercontent.com/33131259/95026864-f204d580-069c-11eb-94f0-776f03843ab5.png)
![image](https://user-images.githubusercontent.com/33131259/95026884-0cd74a00-069d-11eb-9423-68dc372dd1a3.png)

cryptoBeliever:
@OlegMakarenko public key is known when there will be outgoing transaction. This address has only incoming transaction.
Is more like bug in explorer because it shows invalid public key for such address ... For sure it's not 000....0000

cryptoBeliever:
I reported bug for explorer: https://github.com/nemgrouplimited/symbol-explorer/issues/663

OlegMakarenko:
When I enter the accounts public keys from the account tab, the wallet doesn't show any warning and announce HASH_LOCK transaction

cryptoBeliever:
To create multisig you need to have public keys of cosigners.
From address you can't query public key if it's not announced to network.

yilmazbahadir:
> @OlegMakarenko public key is known when you do first outgoing transaction. This address has only incoming transaction. From wallet perspective all is IMO OK.
> @bassemmagdy @yilmazbahadir @evias please confirm.
> 
> Is more like bug in explorer because it shows an invalid public key for such an address ... For sure it's not 000....0000

I believe that's correct.

bassemmagdy:
> @OlegMakarenko public key is known when you do first outgoing transaction. This address has only incoming transaction. From wallet perspective all is IMO OK.
> @bassemmagdy @yilmazbahadir @evias please confirm.
> 
> Is more like bug in explorer because it shows an invalid public key for such an address ... For sure it's not 000....0000

Confirmed.

OlegMakarenko:
@cryptoBeliever @yilmazbahadir @bassemmagdy This should work even if the cosignatory candidate accounts don't have any outgoing transaction. I have just double-check it with SDK and it works. 

[...]

PR #552, opened 2020-10-08
multiple multisig account page issues fix & refactor, explorer uri fix

### LINKED ISSUE #513: Can't convert account to multi-sig with "Recommended" fee
Version: v0.12.1

Scenario: Converting account to 3 of 3 multi-sig doesn't work with the Recommended fee. It's created only hash lock transaction.
When we change fee to Fastest multisig-account is created.

Example (first conversion failed, second with higher fee succeed):
http://explorer-96x.symboldev.network/account/TDMB7MLZZIYTQJ2THNPPKHBCIJGQ3IEMMDO6PJQ

Steve mentioned SDK function that can be used to calculate the correct fee:

The SDK has created a method to calculate the max possible fees for aggregate tx. it takes required number of cosignateries and fee multiplier. https://github.com/nemtech/symbol-sdk-typescript-javascript/blob/9696f4ad65b9026e9c4acbddcbd6613d009148aa/src/model/transaction/AggregateTransaction.ts#L406
FeeMultiplier can be retrieved from network properties, and if required number of cosignateries not know, can use the total cosignateries for a safe pass.

yilmazbahadir:
fixes #513, #528, #540, #543, #550, #551


src/components/AddCosignatoryInput/AddCosignatoryInputTs.ts
```diff
@@ -103,41 +103,12 @@ export class AddCosignatoryInputTs extends Vue {
   }
 
   /**
-   * Emits onAddCosignatory event when cosignatory is input as a public key
+   * Emits onAddCosignatory event when cosignatory is input as an address
    * @private
    * @return {void}
    */
   private async addCosignerFromAddress() {
-    const address = Address.createFromRawAddress(this.cosignatory)
```
evias, line 111: as discussed on slack, this check should not be removed as this _prevents_ end-users from creating invalid transactions.

If this check is not done, and users are allowed to create multisig modifications with unknown cosigners ("unknown" as in "unknown by the network"), those transactions will be *invalid* and the end-users will end up being banned which we really want to avoid.

I suggested to add **steps** and **information** on how to solve the "account must be known" problem.

Also, _converting an unknown account to msig_ shall be fine. Whilst _adding cosigners that are unknown accounts_ is not fine.

evias (dismissed):

src/components/AddCosignatoryInput/AddCosignatoryInputTs.ts
yilmazbahadir: As discussed on the daily since this is an existing problem with Public Key input as well, this will be handled with another issue( https://github.com/nemgrouplimited/symbol-desktop-wallet/issues/561 ), please exclude this concern from this review.

yilmazbahadir: This concern is not valid any more after the 0.96 release. Please see the comment in #561 for the details.

bassemmagdy (approved):

bassemmagdy (approved):

rg911 (approved):

PR #563, opened 2020-10-13
POC of creating a configurable docker image

fboucquez:
This will create a "web" wallet docker image, only used for demonstrations or integration tests atm.

The solution is similar to the explorer image.


fboucquez:
This PR is on hold until we decide if we want to provide images for wallet testing proposes.

rg911 (approved):

PR #564, opened 2020-10-14
Transaction list rest pagination with infinite scroll & conventional pagination

### LINKED ISSUE #532: Full Transaction History using REST Pagination
Current TransactionList component permits only to see latest 100 transactions.

Pagination should be changed to use REST pagination directly as to permit older transaction display.

yilmazbahadir:
fixes #532

### DIFF
--- a/src/components/TransactionList/TransactionListTs.ts
+++ b/src/components/TransactionList/TransactionListTs.ts
@@ -66,6 +66,9 @@ export class TransactionListTs extends Vue {
   })
   address: string
 
+  /**
+   * Number of txs visible in each page
+   */
   @Prop({
     default: 10,
   })
@@ -75,6 +78,14 @@ export class TransactionListTs extends Vue {
     default: 'pagination',
   })
   paginationType: 'pagination' | 'scroll'
+
+  /**
+   * Number of txs to be loaded from repository in each request
+   */
+  @Prop({
+    default: 20,
+  })
+  requestPageSize: number
 
   /**
    * Currently active account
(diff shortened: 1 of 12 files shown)


src/components/TransactionList/TransactionListTs.ts
```diff
@@ -259,11 +295,47 @@ export class TransactionListTs extends Vue {
...
   }
 
+  /**
+   * Loads next page of transactions from repository
+   */
+  public loadMore() {
+    if (!this.currentConfirmedPage.isLastPage) {
+      this.$store.dispatch('transaction/LOAD_TRANSACTIONS', {
+        pageNumber: ++this.currentConfirmedPage.pageNumber,
+        pageSize: 20,
```
rg911, line 315: should the pageSize be hardcoded at all?

rg911 (commented):
Have a look at https://github.com/nemtech/symbol-sdk-typescript-javascript/blob/main/src/infrastructure/paginationStreamer/TransactionPaginationStreamer.ts

if you are using the infinite-scroll here, it would be better to use the `streamer` from the sdk instead.

src/components/TransactionList/TransactionListTs.ts
yilmazbahadir: There are two pagination options implemented (pagination-type: (conventional) pagination | (infinite)scroll). The developer can set it through component properties while reusing the same component, the default type is 'scroll'. To support them both and to comply with the current logic I decided to use repository level pagination instead of pagestreamer(although I liked the streamer idea very much). Hope that makes sense.

yilmazbahadir: > should the pageSize be hardcoded at all?

That is the rest pagination not the visible view-pagesize which is a component property. It could be extracted to a constant though. Because it should be some static optimal value which would reduce the number of calls with the optimal payload size, without putting stress on rest and decreasing the round-trip; and since 10 txs fits to the visible area usually, 20 seemed reasonable to me?

rg911 (approved):

PR #581, opened 2020-10-19
Added version and web warning to wallet

fboucquez:
The current `Vue.config.js` file with capital V has a bunch of configurations that I don't think they are used. The version `vue.config.js` I'm coming here only has the customization I need to inject the web flag and version. I think the wallet was using the default vue config without any customization.

https://cli.vuejs.org/config/#global-cli-config

@OlegMakarenko @rg911 @yilmazbahadir . Do you think the Vue.config.js file was used on linux/mac machines? Maybe on windows as they are case insensitive. 

To run in "Web" Mode

`export WEB=true && npm run dev`

An example of where I would put the banner and version:

image. Please correct the text or the location if you don't like it. It's using absolute position atm. It works but not ideal, pretty old school. 

I wouldn't put a commit tag in the version. Once the version is released, it should be immutable. What happened in the last release is that the version wasn't upgraded automatically so the confusion.  I guess it was not done using Travis or "release" branch.

Fixes https://github.com/nemgrouplimited/symbol-desktop-wallet/issues/580
Fixes https://github.com/nemgrouplimited/symbol-desktop-wallet/issues/566

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@ FROM node:lts-alpine AS builder
 
 WORKDIR /app
 COPY . .
-RUN npm install && npm run build
+RUN export WEB=true && npm install && npm run build
 
 FROM nginx:1.17-alpine AS runner
 COPY --from=builder /app/dist /usr/share/nginx/html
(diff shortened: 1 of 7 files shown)


fboucquez:
@rg911 once this PR is merged we can start publishing alpha/release docker images of the wallet. 

Once this is done, I can replace my private image for the official one and merge https://github.com/nemtech/symbol-bootstrap/pull/43

rg911 (approved):

PR #587, opened 2020-10-22
Recipient address validation

bassemmagdy:
update import statement in private key accounts + fix recipient validation


src/core/validation/CustomValidationRules.ts
```diff
@@ -47,9 +47,22 @@ export class CustomValidationRules {
             validate: (value) => {
                 const isValidAddress = AddressValidator.validate(value);
                 const isValidAlias = AliasValidator.validate(value);
-                if (isValidAddress || isValidAlias) {
+                if (isValidAddress) {
                     return true;
                 }
+                if (isValidAlias) {
+                    const nameSpaceRepository = AppStore.getters['network/repositoryFactory'].createNamespaceRepository();
```
rg911, line 54: Would it be better to move into the store (Namespace store perhaps), so it can be shared by other components?

rg911: BTW, https://github.com/nemgrouplimited/symbol-desktop-wallet/pull/589 I have added the linkedAddress in namespace store as needed for the decrypt message

rg911 (changes requested):
The LinkedAddress has been added in the namespace store by another PR (merged), which return `Address | Null`. Yours returns only boolean which might not be reused by other components. It would be good to use the other store and slightly change your validation logic

rg911 (approved):

PR #592, opened 2020-10-26
Frontend styles

Peersyst:
First frontend stylings.

There's still work to be done, but for the logic team would be nice to have this work to start developing with new styles.


rg911 (commented):
Looks good. 

Some commented out codes left in, are those placeholders?

yilmazbahadir (approved):

PR #600, opened 2020-10-27
Bugfix/allow add child account mnemonic

bassemmagdy:
- removing import private key link from CreateProfile page.

- allow mnemonic root accounts to create child accounts in their profile.


src/views/forms/FormSubAccountCreation/FormSubAccountCreation.vue
```diff
@@ -9,7 +9,7 @@
                             <div class="inputs-container">
                                 <div class="select-container">
                                     <Select v-model="formItems.type" class="select-size select-style">
-                                        <Option v-if="!isPrivateKeyProfile" value="child_account">
+                                        <Option v-if="isPrivateKeyProfile" value="child_account">
```
rg911: Not quite sure about this, can you explain pls?

Logic should be:
- If the current account is an HD root account, then the only option available should be `create child account`
- If the current account is an HD child account or a private key account, then cannot create another child, but can create a separate private key account 

Not sure if it currently supports multiple HD root accounts in a single profile. If it does, then create a root HD account should always be an available option

bassemmagdy: These options are displayed based on type of root account in each profile and not based on each account inside the profile. Since you can't add multiple HD root accounts first accounts defines if it's a private key root profile or HD root profile. And because we agreed on mixing between 2 types of accounts inside profiles, now a private key root profile is allowed to create HD accounts and add more private key accounts and same goes for HD root profiles they have the option of adding more children to their mnemonic or import a private key wallet account.

yilmazbahadir (approved):

PR #603, opened 2020-10-29
transactions: added sent/received filter options, filtering reworked

NikolaiYurchenko:


src/store/Transaction.ts
```diff
@@ -204,56 +269,51 @@ export default {
...
-                                group: TransactionGroup.Unconfirmed,
-                                address: currentSignerAddress,
-                                pageSize: 100,
-                                pageNumber: 1, // not paginating
-                                order: Order.Desc,
-                            }),
-                        ),
-                    );
-                }
+            subscriptions.push(
```
rg911, line 285: Not sure this is 100% right here. The unconfirmed & partial transaction should always be on top of the tx list table. Especially the partial transaction, which requires the current user to cosign. Could you pls double check see if this is working on the list?

src/store/Transaction.ts
```diff
@@ -105,10 +158,11 @@ export default {
     getters: {
         getInitialized: (state: TransactionState) => state.initialized,
         isFetchingTransactions: (state: TransactionState) => state.isFetchingTransactions,
-        confirmedTransactions: (state: TransactionState) => state.confirmedTransactions,
```
rg911, line 108: If the getter get removed, do we still need the mutation below?

src/store/Transaction.ts
```diff
@@ -145,8 +199,20 @@ export default {
         ) => {
             state.partialTransactions = conditionalSort(transactions, transactionComparator);
         },
-        setDisplayedTransactionStatus: (state: TransactionState, displayedTransactionStatus: TransactionGroupState) => {
-            state.displayedTransactionStatus = displayedTransactionStatus;
+        setAllTransactions: (state: TransactionState) => {
+            state.transactions = [...state.confirmedTransactions, ...state.unconfirmedTransactions, ...state.partialTransactions];
+        },
+        filterTransactions: (
```
rg911, line 205: The `ADD_TRANSACTION` or `REMOVE_TRANSACTION` actions below send a refresh signal. How does this mutation handle it? I mean if a websocket listener received an unconfirmed/confirmed transaction (there will be a dispatch from the websocket service I think), will that the `transactions` array get updated?

src/services/TransactionFilterService.ts
```diff
@@ -0,0 +1,94 @@
...
+
+import { Transaction } from 'symbol-sdk';
+import { TransactionFilterOptionsState, TransactionState } from '@/store/Transaction';
+
+/**
+ * TransactionFilter used for filtering transactions by group and sent status.
+ */
+export class TransactionFilterService {
+    /**
+     * Filters transactions depends of selected filter options.
```
rg911, line 25: of => on

src/services/TransactionFilterService.ts
```diff
@@ -0,0 +1,94 @@
...
+            if (filterOptions.isPartialSelected) {
+                result = result.concat(partialTransactions);
+            }
+        }
+
+        return this.filterByRecepient(result, filterOptions, currentSignerAddress);
+    }
+
+    /**
+     * Filters transactions depends of selected sent status filter options.
```
rg911, line 57: of => on

src/store/Transaction.ts
```diff
@@ -44,6 +45,54 @@ export enum TransactionGroupState {
     all = 'all',
 }
 
+export enum TransactionFilterOptions {
+    confirmed = 'isConfirmedSelected',
```
rg911, line 49: just my own opinion.  would using number value + bitwise be better here? So you can skip below the long if / else etc...

https://www.w3schools.com/js/js_bitwise.asp

rg911 (changes requested):
Nice one @NikolaiYurchenko 👍  Looks good.

The main concern is in the store. The refactor bit looks good but seems like it might break some existing function. could you please have a look

src/store/Transaction.ts
```diff
@@ -145,8 +199,20 @@ export default {
         ) => {
             state.partialTransactions = conditionalSort(transactions, transactionComparator);
         },
-        setDisplayedTransactionStatus: (state: TransactionState, displayedTransactionStatus: TransactionGroupState) => {
-            state.displayedTransactionStatus = displayedTransactionStatus;
+        setAllTransactions: (state: TransactionState) => {
+            state.transactions = [...state.confirmedTransactions, ...state.unconfirmedTransactions, ...state.partialTransactions];
```
yilmazbahadir, line 203: I think the correct order would be => partial - unconfirmed - confirmed (Since in the tx list, partial and unconfirmed txs need to always go first.)

src/store/Transaction.ts
yilmazbahadir: Also unconfirmed and partial txs query should be called only if it is the first page. Please notice the `if (pageNumber === 1) ` check above.

NikolaiYurchenko: done

src/store/Transaction.ts
NikolaiYurchenko: yes, we are still using group filtering here

src/services/TransactionFilterService.ts
NikolaiYurchenko: done

src/services/TransactionFilterService.ts
NikolaiYurchenko: done

src/store/Transaction.ts
NikolaiYurchenko: I am using the property value of TransactionFilterOptions as key in TransactionFilterOptionsState

src/store/Transaction.ts
NikolaiYurchenko: done

src/store/Transaction.ts
NikolaiYurchenko: done

NikolaiYurchenko (commented):
comments fixed, ready for qa

rg911:
I have tried to announce a transfer transaction than return back to the transaction list. it is still not working quite right. Once the unconfirmed transaction gets removed (become confirmed), the transaction list changed unexpectedly. Could you please have a look?

rg911 (approved):

PR #605, opened 2020-10-30
Mutlisig graph tree

bassemmagdy:
- Adding multisig Tree to the account page


src/store/Account.ts
```diff
@@ -389,6 +396,20 @@ export default {
                 )
                 .toPromise();
 
+            // sorted array to be represented in multisig tree
+            const getMultisigAccountGraphArraySortedPromise = repositoryFactory
```
yilmazbahadir, line 400: Why are we making the same API call here again? Is it just to get the result as sorted? If so why not we use the previous result and sort it?

bassemmagdy: makes sense

src/store/Account.ts
```diff
@@ -425,6 +433,7 @@ export default {
             commit('multisigAccountsInfo', multisigAccountsInfo);
             commit('currentAccountMultisigInfo', currentAccountMultisigInfo);
             commit('currentSignerMultisigInfo', currentSignerMultisigInfo);
+            // commit('multisigAccountGraph', multisigAccountGraph);
```
yilmazbahadir, line 436: Can we remove this line?

src/components/AccountMultisigGraph/AccountMultisigGraphTs.ts
```diff
@@ -0,0 +1,94 @@
...
+    visible: boolean;
+    public multisigAccountGraphInfo: [];
+    public knownAccounts: AccountModel[];
+
+    get multisigGraphTree(): any[] {
+        if (this.multisigAccountGraphInfo) {
+            const tree = [];
+            this.multisigAccountGraphInfo.map((level: []) => {
+                level.map((entry) => {
+                    // @ts-ignore
```
yilmazbahadir, line 48: We might get rid of the ts-ignores by specifying the correct type for the multisigAccountGraphInfo variable? multisigAccountGraphInfo: MultisigAccountInfo[][] maybe?

src/components/AccountMultisigGraph/AccountMultisigGraphTs.ts
```diff
@@ -0,0 +1,94 @@
...
+                    const selected = this.account.address === entry.accountAddress.address ? true : false;
```
yilmazbahadir, line 49: Do we need a ternary operator here? instead => `const selected = this.account.address === entry.accountAddress.address;`

rg911 (approved):

yilmazbahadir (approved):

PR #610, opened 2020-11-02
Add label in account card telling which type is it

bassemmagdy:
- removing extra components for creating a new account by importing a private key.
-  add label in account card telling which type (seed/private_key) is that account.


src/components/AccountSelectorPanel/AccountSelectorPanel.vue
```diff
@@ -20,6 +20,10 @@
                         <img v-else src="@/views/resources/img/symbol/XYMCoin.png" class="grayed-xym-logo" />
                     </span>
                     <span class="mosaic_name">{{ item.name }}</span>
+                    <span class="account-type"
+                        >{{ item.type === 1 ? `${'('} ${$t('Seed')} ${')'}` : `${'('} ${$t('private_key')} ${')'}` }}
```
yilmazbahadir, line 24: Can we just use the parenthesis instead of ${'('} ?

yilmazbahadir (approved):

PR #613, opened 2020-11-02
update import strategy for new design ui

james-lin71:


Peersyst:
Does it make sense with the Profile import ui PR? I think it conflicts with it. @james-lin71 @rg911

rg911 (approved):

PR #619, opened 2020-11-02
Feature/delete account

bassemmagdy:
- allow user to delete current account from account details page


yilmazbahadir (commented):
Are we showing a confirmation modal before user proceeds? (Something like: You'are about the delete an account. Please make sure you have a backup for future recovery. etc.)

bassemmagdy:
> Are we showing a confirmation modal before user proceeds? (Something like: You'are about the delete an account. Please make sure you have a backup for future recovery. etc.)

@yilmazbahadir we just have a password confirmation modal to make sure the user knows what he's up to but not providing any warning messages inside, the reason behind that is that the button is obvious inside the account details page not even on the card which might light user click it by mistake, meaning it's not a high probability at all that a user could delete his own account without being aware of what he is doing.

yilmazbahadir:
Okay, I'll leave it to the testers and UX designers then. Other than that one, all looks okay to me.

rg911 (approved):

PR #635, opened 2020-11-07
Fix ts errors generated by address-book

Peersyst:
There where some typescript errors generated by the address book feature. When updating package-lock.json file typescript version was downgraded to 3.7.


src/views/modals/ModalContactCreation/ModalContactCreation.vue
```diff
@@ -12,6 +12,7 @@
 </template>
 
 <script lang="ts">
+//@ts-ignore
```
yilmazbahadir, line 15: I believe you shouldn't be needing this here.

rg911 (approved):

PR #637, opened 2020-11-09
Bug/enhancement fixes

bassemmagdy:
- fix minApproval and minRemoval showing wrong values

- block users form adding currently selected account as a cosigner for itself

- add missing translation and remove duplicate ones

- hide multisig graph in non-multisig accounts

- set overlay component position to block transactions forms properly in multisig accounts

- update csv


src/language/en-US.json
```diff
@@ -160,6 +169,7 @@
     "decrypt_message": "Decrypt",
     "define": "Registration",
     "delete_account": "Delete Account",
+    "delete_confirmation_message": "Are you sure do you want to delete?",
```
segfaultxavi, line 172:
```suggestion
    "delete_confirmation_message": "Are you sure you want to delete?",
```

src/language/en-US.json
```diff
@@ -721,19 +735,7 @@
...
-    "form_label_new_contact_name": "New contact name",
-    "form_label_new_contact_address": "New contact address",
-    "delete_confirmation_message": "Are you sure do you want to delete?",
-    "tab_accounts_accounts": "My accounts",
-    "tab_accounts_addressbook": "My contacts",
-    "select_contact": "Select contact",
-    "contact_information": "Contact information"
+    "node": "Node",
+    "to_symbol": "to Symbol",
+    "program_description": "Based on the design of the Symbol blockchain chain, the open source Symbol digital asset system, it is convenient to access the symbol network, create accounts, manage assets and more operations..."
```
segfaultxavi, line 740:
```suggestion
    "program_description": "The most convenient way to access the Symbol blockchain, the digital assets management system. Use it to create accounts, manage tokens and more!"
```

How does this string relate to `program_description[1,2,3]`? It will be harder to maintain 2 different versions!

segfaultxavi (changes requested):

src/language/en-US.json
bassemmagdy: this is not currently used in any page inside app and probably will be deleted

rg911 (approved):

src/language/en-US.json
segfaultxavi: Haha, OK, good to see it removed then. It would be nice to have automated checks in place to ensure that all strings are in use so translator resources are not wasted.

PR #655, opened 2020-11-10
Changed to a static harvesting node list, misc. styling issues fixed

yilmazbahadir:


rg911 (approved):
will need to separate the persistent harvesting out from the key links aggregate transaction. but will work on a separate pr off this

PR #684, opened 2020-11-18
Meta data key

bassemmagdy:
- Removed key from metadata creation form and auto-generate random 8 bytes key in background.

- Peersyst UI updates


rg911:
Closed as new PR to be created

PR #716, opened 2020-11-20
NodeSelector filtering nodes with roleType peer and minor styling fixes

yilmazbahadir:


src/store/Network.ts
```diff
@@ -66,30 +67,37 @@ const staticPeerNodes: NodeInfo[] = [
     {
         nodePublicKey: 'BE60BE426872B3CB46FE2C9BAA521731EA52C0D57E004FC7C84293887AC3BAD0',
         host: 'beacon-01.ap-northeast-1.0.10.0.x.symboldev.network',
+        roles: [RoleType.PeerNode],
```
rg911, line 70: will this get filtered out anyway?

I guess we should keep these static nodes until testnet refresh.

yilmazbahadir: No, it won't. We are keeping the following role types: [PeerNode, VotingNode] and filtering out the rest. So static nodes will be there.
Screenshot 2020-11-20 at 17 45 28

rg911 (approved):

PR #768, opened 2020-12-03
show notification in case mnemonic not valid before going to next ste…

### LINKED ISSUE #710: Bug - Mnuemonics order changed - error message is not appearing
While creating an account, via mnemonics , if I change the order, error message need to be displayed ,it is not displaying .screenshot attached 
image
image

bassemmagdy:
fixes #710


yilmazbahadir (approved):

src/components/MnemonicVerification/MnemonicVerificationTs.ts
```diff
@@ -87,18 +86,41 @@ export class MnemonicVerificationTs extends Vue {
         const rebuilt = this.selectedWordIndexes.map((i) => this.shuffledWords[i]).join(' ');
 
         // - origin words list does not match
+        if (origin !== rebuilt) {
+            this.mnemonicCheckerNotification(origin, rebuilt);
```
rg911, line 90: Dup codes. It can be moved before the check and keep only 1 call I guess

rg911 (changes requested):

src/components/MnemonicVerification/MnemonicVerificationTs.ts
bassemmagdy: Should be resolved

Issue #787, opened 2020-12-09
Tips - Mneumonics for password backup is always displayed as invalid


surekabpm:
While generating the account , at the verification stage for the back up mneumonics Tips is displayed as "Your mnemonic words backup is invalid. Please, carefully select the words in the correct order.".
I have not yet entered any words, but is this display right? kindly review 
![image](https://user-images.githubusercontent.com/49143210/101596361-ecb86400-3a1a-11eb-8f7c-3204ace223d6.png)
@bassemmagdy

cryptoBeliever:
Fixed by changeing tooltip.

Issue #760, opened 2020-12-01
Unable to convert account to multi-signature


OlegMakarenko:
The wallet announce **HASH_LOCK** without **AGGREGATE_BONDED**
![image](https://user-images.githubusercontent.com/33131259/100784569-0c96c900-3418-11eb-8a58-06f113534a42.png)
![image](https://user-images.githubusercontent.com/33131259/100784598-17515e00-3418-11eb-9e0a-acaf9bdcdfba.png)
![image](https://user-images.githubusercontent.com/33131259/100784624-220bf300-3418-11eb-8c47-7f2e5c57f277.png)
![image](https://user-images.githubusercontent.com/33131259/100784638-26d0a700-3418-11eb-999b-5c6ecb3157cf.png)

cryptoBeliever:
@OlegMakarenko what not did you used?

OlegMakarenko:
@cryptoBeliever sorry, I don't understand your question

cryptoBeliever:
@OlegMakarenko sorry. Autocorrect. 
Which one node was selected when you tried convert to multisig? I'm asking becauase there was problems with couple of nodes.

OlegMakarenko:
@cryptoBeliever I've just tried one more time and it works. I don't remeber I changed a node 
![image](https://user-images.githubusercontent.com/33131259/101035456-6b2c8600-3583-11eb-8c81-80bd16eab0b5.png)

bassemmagdy:
@OlegMakarenko please confirm and provide reproduction steps if this's still happening. I couldn't reproduce it.

OlegMakarenko:
@bassemmagdy 
1. Generate 3 seed accounts (let's call them **s1**, **s2**, **ms**)
2. Send 500 XYM to each of them
3. Select **ms** account and go to the Multisig page
4. Add **s1** and **s2** as cosignatories
5. Set minApproval = 2. Set minRemoval = 1
6. Confirm
7. Send

cryptoBeliever:
@Eugenity @NikolaiB please check it too if possible.

NikolaiB:
@cryptoBeliever I couldn't test it yesterday due to node issue. When the node issue is fixed I will test it.

NikolaiB:
Fixed

Issue #741, opened 2020-11-25
[Bug][Multisig account] Can not Remove Cosignatory 


NikolaiB:
WTR:
1. Create multisig account from 2 separate desktop wallets (not child account)
2. Go to Multisig page and try to Removed Cosignatory a few times
3. Go to Dashboard

Result: You can see that found was locked a few times but cosignatory still not removed 

![image](https://user-images.githubusercontent.com/11311924/100236759-49664a00-2f58-11eb-85f2-d99e80ea0d07.png)

bassemmagdy:
related to #760

NikolaiB:
Fixed

PR #826, opened 2020-12-16
fixes multisig account bugs

bassemmagdy:
fixes

-  cannot remove a cosignatory #760 
- cannot convert account to multisig #741
- mnemonic tip is always displayed as invalid #787


src/store/Account.ts
```diff
@@ -474,9 +474,13 @@ export default {
             commit('accountsInfo', accountsInfo);
 
             // read signer info to get public key
-            const signerModel = accountsInfo.find((w) => w.address.equals(currentSignerAddress));
+            const signerModel = knownAccounts.find((w) => w.address === currentSignerAddress.plain());
```
rg911, line 477: why use knownAccounts here?

bassemmagdy: Because accountsInfo gets updated in storage on account selection. So it might cause **Failure_Core_Zero_Public_Key** or sending Hash_Lock transactions without an Aggregate one following its confirmation while trying to interact with accounts in same profile that might be virgin or accounts that have received some funds already but, still didn't get selected yet to update their accountsInfo data in storage.

rg911 (approved):

PR #831, opened 2020-12-16
account restrictions: ui bugs fix

NikolaiYurchenko:


bassemmagdy (changes requested):
Great Work @NikolaiYurchenko! It just misses the part for hiding the mosaic restriction deletion window after deletion confirmation, the same way as you already did for the Address restrection deletion modal.

bassemmagdy (approved):

Issue #838, opened 2020-12-18
Under Multisig account co-signers are displayed as "(Multisig)" in "Multisig accounts" filter


Eugenity:
**Steps to reproduce:**

1. Create a Multisig account
2. Go to Dashboard or Accounts/restrictions page, Mosaics, Namespaces
3. Click on the "Multisig accounts" filter

**Expected result:** Co-signers are indicated as "(Co-signer)" or similar
**Actual result:** Co-signers are displayed as "(Multisig)" in the filter

[Screenshot](https://share.getcloudapp.com/4guOZ4Nm)

cryptoBeliever:
@Eugenity please verify.

Eugenity:
Verified

PR #841, opened 2020-12-20
Update rssmix news url to use HTTPS

bassemmagdy:
1. Changing news url to use this [one](https://dev-vibes.github.io/symbol-news/) instead of rssmix to fetch news over HTTPS  to avoid MiTM attacks and responses manipulation.

2. fixes multi-sig issue #838.


src/services/MultisigService.ts
```diff
@@ -94,9 +94,9 @@ export class MultisigService {
                 return (addressesFromNextLevel = addressesFromNextLevel.concat([
                     {
                         address: term.accountAddress,
-                        multisig: true,
+                        multisig: !!term.cosignatoryAddresses.length ? true : false,
```
NikolaiYurchenko, line 97: I think the ternary operator is redundant. We  can just use `multisig: !!term.cosignatoryAddresses.length`

rg911 (approved):

rg911 (approved):

PR #850, opened 2020-12-23
qr code message encryption fixed, import transaction uri modal fixed

NikolaiYurchenko:


bassemmagdy:
@NikolaiYurchenko I think the code needs to be linted.

rg911 (approved):

Issue #808, opened 2020-12-10
Restrictions - Not possible to lift "account operation restriction" restriction


cryptoBeliever:
Steps:

1. Add restriction:
Type: Account operation restriction
Restriction Additions: Account operation restriction
Restriction Flag: AllowOutgoingTransactionType

**Now all should be blocked except lifting restriction. And indeed I got Failure_RestrictionAccount_Operation_Type_Prohibited when trying send Transfer Transaction**

2. Try remove restriction. Not possible with error **Failure_RestrictionAccount_Invalid_Modification**

What is even stranger I'm able to add another operation restriction and I'm able to remove it.

yilmazbahadir:
Then we should put a restriction to "account operation restriction" :)

cryptoBeliever:
Fixed by removing "Allow" from "Type" and "Account operation restrictioin" from "Operation type" selectboxes. Network behaviour stays unchanged.
Thanks @bassemmagdy :+1:

Issue #811, opened 2020-12-11
Restrictions - Operaton restrictions allowing not possible


cryptoBeliever:
Steps:
1. Go to operation restrictions
2. Create new operation restriction
Direction: Outgoing
Type: Allow
Operation type: **Any except Account operation restriction**
3. Submit

Result: Failure_RestrictionAccount_Invalid_Modification

This is because allowing any would cause blocking others and this implies not possible to lift such restriction.
Imo Allow type should be blocked here. It works only for "Account operation restriction" but there is problem with removing it as described in https://github.com/nemgrouplimited/symbol-desktop-wallet/issues/808

cryptoBeliever:
Fixed by removing "Allow" from "Type" selectbox. Network behaviour stays unchanged.
Thanks @bassemmagdy :+1:

Issue #807, opened 2020-12-10
Restrictions - not possible to to restrict operation type "Account operation restriction" :)


cryptoBeliever:
Steps:
1. Go to the account
2. Go to restrictions
3. Choose "Operation restrictions"
4. Choose Type "Block" 
5. Choose Operation Type "Account operation restriction"

Result: I got **Failure_RestrictionAccount_Invalid_Modification** error

Is this by design in Symbol blockchain (transaction is blocked because not announced yet restriction)?

cryptoBeliever:
Fixed by removing "Account operation restriction" from "Operation Type" selectbox.
Thanks @bassemmagdy :+1:

PR #853, opened 2020-12-24
fix bug cannot lift restriction

bassemmagdy:
- Issues Fixed:
#808 #807 #811

- After discussion with @cryptoBeliever and @yilmazbahadir on account restrictions issues on wallet we concluded that the solution to tackle them so far on client-side could happen by modifying couple of things : 

1.  Eliminate Account Operation Restriction from operation types, so users are not able to use it either to block or allow that specific type.

2. All the other operation restriction types only block would be allowed.

### DIFF
--- a/src/views/forms/FormAccountRestrictionTransaction/FormAccountRestrictionTransaction.vue
+++ b/src/views/forms/FormAccountRestrictionTransaction/FormAccountRestrictionTransaction.vue
@@ -8,7 +8,10 @@
 
                     <RestrictionDirectionInput v-model="formItems.direction" :disabled="isDeleteMode || directionDisabled" />
 
-                    <RestrictionTypeInput v-model="formItems.blockType" :disabled="isDeleteMode || transactionTypeDisabled" />
+                    <RestrictionTypeInput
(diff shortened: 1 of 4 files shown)


src/views/forms/FormAccountRestrictionTransaction/FormAccountRestrictionTransaction.vue
```diff
@@ -8,7 +8,10 @@
 
                     <RestrictionDirectionInput v-model="formItems.direction" :disabled="isDeleteMode || directionDisabled" />
 
-                    <RestrictionTypeInput v-model="formItems.blockType" :disabled="isDeleteMode || transactionTypeDisabled" />
+                    <RestrictionTypeInput
+                        v-model="formItems.blockType"
+                        :disabled="isDeleteMode || transactionTypeDisabled || restrictionTxType === 'TRANSACTION_TYPE'"
```
yilmazbahadir, line 13: Should we move `restrictionTxType === 'TRANSACTION_TYPE'` into `transactionTypeDisabled` method ?

yilmazbahadir (approved):

PR #870, opened 2021-01-05
fix multisig page is partially disabled

### LINKED ISSUE #862: Part of the Multisig page is not blocked under Multisig account
**Steps to reproduce:**

1. Go to the Multisig page under the Multisig account
2. Click on the bin buttons for a co-signer added or on the right side of the Send button

**Actual result:** Part of the Multisig page for a Multisig account is not blocked for interaction
**Expected result:** All the Multisig page is blocked for interaction under Multisig account

[Video](https://share.getcloudapp.com/L1uOv48e)

bassemmagdy:
fixes #862

### DIFF
--- a/src/components/DisabledFormOverlay/DisabledFormOverlay.less
+++ b/src/components/DisabledFormOverlay/DisabledFormOverlay.less
@@ -16,7 +16,8 @@
 .multisig_ban_container {
     font-size: 16px !important;
     .is_multisig {
-        width: 12rem;
+        width: 100%;
+        min-width: 12rem;
         height: 100%;
         position: absolute;
         background-color: white;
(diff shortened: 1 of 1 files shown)


src/components/DisabledFormOverlay/DisabledFormOverlay.less
```diff
@@ -16,7 +16,7 @@
 .multisig_ban_container {
     font-size: 16px !important;
     .is_multisig {
-        width: 100%;
+        width: 12rem;
```
yilmazbahadir, line 19: we shouldn't be fixing the overlay issue by setting a fixed width imo. Do you think there might be another way to  fix this while keeping the width as 100% ?

rg911 (approved):

PR #882, opened 2021-01-06
Reduced Mosaic amount to 1

OlegLialin:


bassemmagdy:
@OlegLialin This breaks the divisibility of mosaics as the transaction details form is now never showing the right mosaic amount! for example using the maximum allowed value it shows 9,000,000,000 while the maximum value is 9000000000000000! this is not a special case try to use any amount and check the details amount before sending the transaction

yilmazbahadir:
@bassemmagdy @rg911 @OlegLialin Should we close this PR as the change doesn't look correct?

PR #888, opened 2021-01-12
use fetch for get requests instead of axios

bassemmagdy:


src/services/CommunityService.ts
```diff
@@ -30,8 +29,14 @@ const request = async (): Promise<string> => {
         feedUrl = '/nemflash';
     }
     // execute request
-    const response = await axios.get(feedUrl, { params: {} });
-    return response.data;
+    const response = await fetch(feedUrl, {
+        method: 'GET',
+    })
+        .then((response) => response.text())
```
fboucquez, line 35: I'm not sure if axios's data is a text or a parsed json. Please double-check this. A unit test would do...

src/services/CommunityService.ts
```diff
@@ -30,8 +29,14 @@ const request = async (): Promise<string> => {
...
+        .then((str) => {
+            return str;
```
fboucquez, line 37: Doesn't do anything right?

src/services/CommunityService.ts
bassemmagdy: Already checked axios response before changing it to use fetch and it was returning text, not json.

src/services/CommunityService.ts
bassemmagdy: Yes, and it was here because I was trying to create an XML document before creating a string out of it. Just updated it

yilmazbahadir (approved):

yilmazbahadir (approved):

PR #916, opened 2021-01-16
Reworked node selector layout

myxmaster:
Now looks like this:

image


Peersyst (changes requested):
Can you center as well the top titles? Selected network and node?

Thank you!

myxmaster:
That was on purpose. I think the readability is a lot better like this because when you center it, the distance between label (Current Network/Node) and selected network/node url is way too big.

rg911 (approved):

PR #945, opened 2021-01-20
Several address handling fixes

myxmaster:
- Now copying pretty account address to clipboard instead of plain
- Removed spaces before and after the address in account details
- Now saving contact addresses as plain and therefore preventing creation of duplicate contacts
- Always display pretty contact addresses instead of saved one

### DIFF
--- a/src/components/AccountAddressDisplay/AccountAddressDisplay.vue
+++ b/src/components/AccountAddressDisplay/AccountAddressDisplay.vue
@@ -1,10 +1,8 @@
 <template>
     <div class="account-detail-row-3cols">
         <span class="label">{{ $t('account_address') }}:</span>
-        <div class="value accountAddress">
-            {{ getPrettyAddress() }}
-        </div>
-        <ButtonCopyToClipboard v-model="address">
+        <div class="value accountAddress">{{ prettyAddress }}</div>
+        <ButtonCopyToClipboard v-model="prettyAddress">
             <img src="@/views/resources/img/account/cloneIcon.svg" class="copy-icon" />
         </ButtonCopyToClipboard>
     </div>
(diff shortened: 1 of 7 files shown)


myxmaster:
@cryptoBeliever told me that most exchanges don't work with pretty addresses (with hyphens), so I guess the first point (_Now copying pretty account address to clipboard instead of plain_) should not be done.

rg911:
> @cryptoBeliever told me that most exchanges don't work with pretty addresses (with hyphens), so I guess the first point (_Now copying pretty account address to clipboard instead of plain_) should not be done.
> 
> On the other hand we **do** work with pretty addresses (for good reasons!). So users could get the pretty addresses from other places (if they don't use the copy-button) and then they still have this problem with the exchange... It could be extra confusing if they sometimes get the pretty address and sometimes the plain address.

The pretty format is for displaying only. All address operation should be `plain`. Buth 'pretty' helps when 2 addresses come with similar strings. Personally I would prefer showing pretty (with copying the plain format). It is the same as NIS1 nano wallet. 
@cryptoBeliever @bassemmagdy @AdriaCarrera ?

cryptoBeliever:
> > @cryptoBeliever told me that most exchanges don't work with pretty addresses (with hyphens), so I guess the first point (_Now copying pretty account address to clipboard instead of plain_) should not be done.
> > On the other hand we **do** work with pretty addresses (for good reasons!). So users could get the pretty addresses from other places (if they don't use the copy-button) and then they still have this problem with the exchange... It could be extra confusing if they sometimes get the pretty address and sometimes the plain address.
> 
> The pretty format is for displaying only. All address operations should be `plain`. But 'pretty' help when 2 addresses come with similar strings. Personally, I would prefer showing pretty (with copying the plain format). It is the same as NIS1 nano wallet.
> @cryptoBeliever @bassemmagdy @AdriaCarrera ?

@rg911 yes showing pretty and copying plain format is fine for me. In NIS1 it work in this way (if user using copy button)

myxmaster:
Alright, you convinced me.

rg911 (approved):

rg911:
@myxmaster  any chance you could have a look at the conflicts?

myxmaster:
Conflicts resolved

rg911 (approved):

PR #950, opened 2021-01-21
Update ledger intergration

TriPhan96:
- Create account/subaccount and login with Ledger Nano S devices
- Sign simple and aggregate transaction
- Allow delegated harvesting for Ledger account
- Work with Symbol BOLOS app v0.0.4 or higher


package.json
```diff
@@ -87,10 +94,11 @@
         "babel-preset-env": "^1.7.0",
         "bufferutil": "^4.0.1",
         "cross-env": "^5.1.1",
-        "electron": "11.1.0",
+        "electron": "^8.2.0",
```
bassemmagdy, line 97: @TriPhan96 electron version should be 11.1.0

src/core/utils/Ledger.ts
```diff
@@ -0,0 +1,236 @@
...
+
+        while (offset !== rawTx.length) {
+            const maxChunkSize = offset === 0 ? MAX_CHUNK_SIZE - 1 - bipPath.length * 4 : MAX_CHUNK_SIZE;
+            const chunkSize = offset + maxChunkSize > rawTx.length ? rawTx.length - offset : maxChunkSize;
+            // APDU fields configuration
+            const apdu = {
+                cla: CLA_FIELD,
+                ins: TX_INS_FIELD,
+                p1: offset === 0 ? (chunkSize < maxChunkSize ? 0x00 : 0x80) : chunkSize < maxChunkSize ? 0x01 : 0x81,
+                p2: curveMask | (chainCode ? 0x01 : 0x00),
```
bassemmagdy, line 211: isn't `chainCode` always false? which means that this term will always evaluate to 0x00?

src/views/forms/FormProfileCreation/FormProfileCreationTs.ts
```diff
@@ -150,13 +210,33 @@ export class FormProfileCreationTs extends Vue {
...
+            this.$store.dispatch('notification/ADD_SUCCESS', NotificationType.OPERATION_SUCCESS);
 
-        // flush and continue
-        this.$router.push({ name: this.nextPage });
+            // flush and continue
+            this.$router.push({ name: this.nextPage });
+        } else {
+            // execute store actions
+            this.$store.dispatch('profile/SET_CURRENT_PROFILE', account);
+            this.$store.dispatch('temporary/SET_PASSWORD', this.formItems.password);
```
bassemmagdy, line 224: These two lines for executing store actions are redundant, take both of them out of conditions and use them once.

src/core/utils/Ledger.ts
TriPhan96: Okay, i'm going update to remove the hard code value, but will change to a parameter cause chain code may be require for verification later

src/views/forms/FormProfileCreation/FormProfileCreationTs.ts
TriPhan96: Okay, i will update it

package.json
TriPhan96: Hi @bassemmagdy, we have upgrade the Electron package to [version 11.1.0](https://github.com/FDS-Ledger/symbol-desktop-wallet/blob/fb1ac37cb2424fa239eabd7d63529f30482e41e0/package.json#L101). Please have a look at this PR, thank you.

src/core/utils/Ledger.ts
```diff
@@ -0,0 +1,230 @@
...
+        const MAX_CHUNK_SIZE = 255;
+        const CONTINUE_SENDING = '0x9000';
+
+        const curveMask = 0x80; // use Curve25519
+        const bipPath = BIPPath.fromString(path).toPathArray();
+        const apduArray = [];
+        let offset = 0;
+
+        while (offset !== rawTx.length) {
+            const maxChunkSize = offset === 0 ? MAX_CHUNK_SIZE - 1 - bipPath.length * 4 : MAX_CHUNK_SIZE;
```
bassemmagdy, line 198: @TriPhan96 `offset` always 0? meaning no need for ternary operator?

tri-vo-fds: Hi, thanks @bassemmagdy.  Since the _offset_ variable value will be changed in the while loop, so the conditional operators with it make sense

bassemmagdy (approved):

Issue #670, opened 2020-11-14
Format of certain properties of an address


garm01:
The wallet should be easy to understand for the average user. After all, these are the people who use it.
This also applies to the "Importance" of an address. It should be displayed as in the explorer. 
e.g. "0.001352%"

![1Importance](https://user-images.githubusercontent.com/46584115/99153148-0201d300-26a7-11eb-83f0-af6292b3bd70.png)

garm01:
and look at here: 

![2Importance](https://user-images.githubusercontent.com/46584115/99153856-9110ea00-26ab-11eb-9563-9fdd8cd7b232.png)

cryptoBeliever:
Interesting. I guess it depends on localization because I can see:

![image](https://user-images.githubusercontent.com/17684281/99313005-23a5ba80-285f-11eb-9e31-8aad952ff295.png)

garm01:
Not sure, I used " [0.13.4][v0.13.4] - 11-Nov-2020 "

garm01:
The first picture :
#locale 
LANG=de_DE.UTF-8
LC_CTYPE="de_DE.UTF-8"
LC_NUMERIC="de_DE.UTF-8"
LC_TIME="de_DE.UTF-8"
LC_COLLATE="de_DE.UTF-8"
LC_MONETARY="de_DE.UTF-8"
LC_MESSAGES="de_DE.UTF-8"
LC_PAPER="de_DE.UTF-8"
LC_NAME="de_DE.UTF-8"
LC_ADDRESS="de_DE.UTF-8"
LC_TELEPHONE="de_DE.UTF-8"
LC_MEASUREMENT="de_DE.UTF-8"
LC_IDENTIFICATION="de_DE.UTF-8"
LC_ALL=
                                       -----
and here:
#locale
LANG=en_GB.UTF-8
LC_CTYPE="en_GB.UTF-8"
LC_NUMERIC="en_GB.UTF-8"
LC_TIME="en_GB.UTF-8"
LC_COLLATE="en_GB.UTF-8"
LC_MONETARY="en_GB.UTF-8"
LC_MESSAGES="en_GB.UTF-8"
LC_PAPER="en_GB.UTF-8"
LC_NAME="en_GB.UTF-8"
LC_ADDRESS="en_GB.UTF-8"
LC_TELEPHONE="en_GB.UTF-8"
LC_MEASUREMENT="en_GB.UTF-8"
[...]

PR #957, opened 2021-01-23
Fix importance score display

myxmaster:
This fixes the importance score display (issue #670).

Therefore the totalChainImportance had to be added to the NetworkConfigurationModel with a default value. I created this pull request as a draft because I have no idea what default value should be used here. Actually I don't know why default values are used at all as it would lead to displaying wrong values. I set it to 1 which is maybe not what you want.

### DIFF
--- a/src/config/NetworkConfig.ts
+++ b/src/config/NetworkConfig.ts
@@ -38,6 +38,7 @@ export interface NetworkConfigurationDefaults {
     maxMosaicDivisibility: number;
     maxMessageSize: number;
     epochAdjustment: number;
+    totalChainImportance: number;
 }
 
 export interface NetworkConfig {
@@ -72,6 +73,7 @@ const defaultNetworkConfig: NetworkConfig = {
         harvestingMosaicId: '5F160D7851F3CB30',
         defaultDynamicFeeMultiplier: 1000,
         epochAdjustment: 1573430400,
+        totalChainImportance: 1,
     },
     nodes: [
         { friendlyName: 'API North East 1', roles: 2, url: 'http://api-01.ap-northeast-1.testnet.symboldev.network:3000' },
--- a/src/config/NetworkConfig.ts
+++ b/src/config/NetworkConfig.ts
@@ -73,7 +73,7 @@ const defaultNetworkConfig: NetworkConfig = {
         harvestingMosaicId: '5F160D7851F3CB30',
         defaultDynamicFeeMultiplier: 1000,
         epochAdjustment: 1573430400,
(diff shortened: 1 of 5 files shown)


src/config/NetworkConfig.ts
```diff
@@ -72,6 +73,7 @@ const defaultNetworkConfig: NetworkConfig = {
         harvestingMosaicId: '5F160D7851F3CB30',
         defaultDynamicFeeMultiplier: 1000,
         epochAdjustment: 1573430400,
+        totalChainImportance: 1,
```
rg911, line 76: you can use the default value in the testnet, which will be the same as main net after launch

http://api-01.us-west-1.testnet.symboldev.network:3000/network/properties

rg911 (commented):
Looks good to me 👍

myxmaster:
After more thoughts (and a talk to a core dev) about it, I think we shouldn't use a default value at all for this, but display "?" or "N/A" or something instead, if totalChainImportance can't be retrieved. 

Reason:
The total chain importance changes with the total XYM supply. If we use a default value for the calculation of the relative node importance (in case totalChainImportance can't be retrieved), the resulting relative node importance seems correct, but is in fact wrong.

If you agree, I'll change the commit accordingly.

rg911:
> After more thoughts (and a talk to a core dev) about it, I think we shouldn't use a default value at all for this, but display "?" or "N/A" or something instead, if totalChainImportance can't be retrieved.
> 
> Reason:
> The total chain importance changes with the total XYM supply. If we use a default value for the calculation of the relative node importance (in case totalChainImportance can't be retrieved), the resulting relative node importance seems correct, but is in fact wrong.
> 
> If you agree, I'll change the commit accordingly.

The default *IS* the default value :) It will be set in the node configuration when the main launched (test net current has the same settings as main). so I think we are safe to use default here

myxmaster:
The default value is not the correct value though. It is slightly different than the real (correct) value. So what happens then, in case the real total chain importance can't be retrieved from the network, is that the displayed (relative node) importance is close to the real value, but not 100 % correct. The user then has no chance to understand that it is not correct.

So my conclusion is: In such case there should be a clear sign that there was a problem and the score couldn't be calculated.

rg911:
@myxmaster has this been completed? or still in draft?

myxmaster:
No, not yet. But if you agree I'll change it so that there is a "N/A" instead of wrong importance score (in case total chain importance couldnt be retrieved). And then I'll ping you when it's done.

rg911 (approved):

PR #961, opened 2021-01-23
Use document's mouseover event for entropy generation

myxmaster:
When "generating a mnemonic" currently the mouseover event of the GenerateMnemonic component is used. Now using the document's mouseover event for more convenience and maybe even more randomness.


rg911 (commented):
can DOM be accessed in electron? or apps perhaps?

myxmaster:
I tried to use the document.onmousemove event in an electron app and it works.

What do you mean by this?

> or apps perhaps?

rg911 (approved):

Issue #965, opened 2021-01-25
Delegated Harvesting Node Url Selection Problem


ryo-opening-line:
I found the issue related to selecting a delegated harvesting node in the desktop wallet as follows:

1. Select a node from list
![image](https://user-images.githubusercontent.com/28587337/105664887-21c03d80-5f19-11eb-93bc-0baaa16e1711.png)

2. Push Start and cancel the Sign transaction(s) dialog
![image](https://user-images.githubusercontent.com/28587337/105665038-6f3caa80-5f19-11eb-90ef-67230c855e73.png)

3. Type another node name 
![image](https://user-images.githubusercontent.com/28587337/105665088-8da2a600-5f19-11eb-815a-ac948a1100a0.png)

4. Start and confirm the Sign transaction(s) dialog
![image](https://user-images.githubusercontent.com/28587337/105665146-b4f97300-5f19-11eb-88ed-1a8659e22134.png)

You can see that the information of the previously selected node is still there.

Could you check it, please?

Thanks!

cryptoBeliever:
@bassemmagdy could you check?

cryptoBeliever:
Harvesting was reworked (still in tests). This bug was fixed (you can check in live build http://desktop-build.symboldev.com/)

Issue #952, opened 2021-01-21
Priority - Delegated harvesting to persist linked Keys


rg911:
At the moment Delegated harvesting registration won't save up the linked keys. It is proposed that:

#### Delegated harvesting improvement
1. Separate key link transactions and persistent harvesting transaction's creation & announcement
2. Step 1 (currently the `start` button): Generate Remote, VRF key pairs, create and announce 3 (Remote, VRF, Node) key link transactions in an aggregate complete. Encrypt the **Private** key for `Remote` and ` VRF` keys, save the encrypted keys into the storage. 
3. Step 2 (current the `activate`):  Load the vrf private key from local storage, create and announce the persistent harvesting transaction. 
4. All other main delegated harvesting function stays the same.

#### Disaply linked keys
1. Add a new section in account info to show Linked **Public** keys. (accountInfo store)
2. Link any saved `Remote` and `VRF` private keys to the above public keys. 
3. Ability to display the private key (password needed), for those who need them to setup their own nodes.

segfaultxavi:
Will this remove the need for the two-step process to enable Delegated Harvesting?

PR #987, opened 2021-01-28
update delegated harvesting

bassemmagdy:
- reworked delegated harvesting #952 #965
- seperate linked keys transactions from persistent del request
- add new section to show linked public and private keys
- add `link/unlink` action for each linked public key (node - vrf - remote)
- update key link transaction to include only unlinked keys


rg911 (changes requested):
#### Issues
1. If the keys are linked and unlinked,  the private keys are not removed from storage. 
2. Unlinked the keys then relink, the page is asking a swap. should be: activate. (probably related to above)
3. on node selector, there could be peer nodes that don't have harvesting enabled. So in node select event, there should be 2 checks: 1) `node/info`, check nodePublicKey, 2) if nodePublicKey exists, check `node/unlockedaccount`, if unlockedAccount endpoint returns 503 error, it is either harvesting not enabled or API node is down, so in this case, an error message should pop up to ask the user to select different nodes. 

#### Improvements
- Key link section should always show on the page. If there is no key linked, show `unlinked` otherwise show the linked public key. 
- For each link key, there should be a button to allow `Link` / `Unlink` with single transaction. this is for 'remote' harvester to only link remote key and vrf. The `Link Keys` button can stay, for delegated harvest to link all 3 keys in an aggregate.
- Linked Key title to remove `Linked` from the beginning.
- VRF and Remote keys get generated from seeds (HD Wallet). So they will be included in the key backup.  We are also working on a tool to collect the remote, vrf,and voting keys from option, the index used in that tool: 1: VRF, 3: Remote (2: voting, not used in delegated harvesting)

rg911 (changes requested):
1. the remote & vrf key link / unlink should not require a peer node selected. Only Node link does
2. when there is no private key stored (remote & vrf), hide the view button or show 'no private key stored'

rg911 (approved):

PR #989, opened 2021-01-28
Close settings modal dialog on confirm

myxmaster:
When confirming general settings, the dialog now closes after entering the (correct) password.


rg911 (commented):
Not sure if we need to auto close the modal. What if other settings need changing?

myxmaster:
I think its a rather rare case that somebody wants to change general settings (change password e.g.) and then also add a node? So I thought in most cases it is annoying for the user to confirm his changes and then manually close the dialog afterwards. But I guess users don't often change general settings, so either way its not a big deal.

rg911 (approved):

Issue #946, opened 2021-01-20
Application can be used without accepting terms and conditions


myxmaster:
Currently the modal dialog which prompts to accept the terms and conditions can be closed without accepting by using the cross on the top right corner or by clicking on the grayed out background. It is also possible, after clicking on the terms and conditions link, to navigate to all pages except Home. The terms and conditions should probably also be shown in a modal dialog.

cryptoBeliever:
Fixed.

cryptoBeliever:
Still I'm able to use wallet without accepted terms.
1. Click terms details
2. Go to transaction view.
Pop-up shows again only if I click home.

Maybe it should be enforced on profile creation/importing?

cryptoBeliever:
Current solution prevents creating profile without accepting conditions. Not sure if those are final version and if not should be available "outside" wallet.

PR #1002, opened 2021-02-01
Application can be used without accepting terms and conditions #946

james-lin71:


rg911 (changes requested):
why create the terms in the wallet modal? These supposed to be an external link controlled by the legal and marketing team

james-lin71:
Oh, I didn't care about  for legal and marketing team, will make change to be support with external link.

rg911 (commented):
@bassemmagdy @OlegMakarenko 

I believe this has been fixed already by @OlegMakarenko ?

james-lin71:
There were additional exception that users can navigate to the some other pages like 'namespaces', 'mosaic' and etc.
The terms & conditions accept modal was restricting user only on home page that shows transactions history.
So, some suggestion was moving the terms & conditions accept logic to the create/import profile logic, and I integrated this on finalize of create/import profile page to make sure users must accept terms and conditions before their use.

bassemmagdy (approved):

rg911 (approved):

Issue #940, opened 2021-01-19
The selection dialog for the fees for aggregate transactions is not completely displayed.


garm01:
![fee2](https://user-images.githubusercontent.com/46584115/105097016-37aca900-5aa8-11eb-8020-c61ae541a39d.png)

cryptoBeliever:
Fixed.

PR #1008, opened 2021-02-01
Reworked and fixed aggregate transactions layout and logic

myxmaster:
- Reworked layout
  - More space for transactions list
  - Now using navigation tabs for transaction types
  - Editing transactions now via modal dialog
  - Fee selector dropdown now opens to the top so that it is no longer cut off (#940 fixed)
  - Removed useless whitespace in confirmation dialog and added separators between transactions
- Aggregate transactions storage now saves and returns object clones so that components that modify an object do not edit the stored object
- Deleting a transaction no longer requires password (creating one doesn't either)

Questions:
- Why can't you encrypt messages in an aggregate transaction?
- Why does `emitToAggregate()` update mosaics and signerPublicKey? Why not update it in `triggerChange()`? The logic of ModalTransactionEdit would be simpler if `emitToAggregate()` would not have to be called.

grafik

grafik

grafik


rg911 (approved):

garm01:
Under "Home" and "Accounts" the smaller field is on the right side.
Under "Harvesting" and "Aggregate" it´s on the left side.
When the user switches directly e.g. from "Accounts" to "Aggregate", it's a bit strange.
Maybe we should make it consistent across the whole wallet ?

myxmaster:
That is true, but those are very different environments and therefore you can't really compare it. I like the dashboard and the contact section how it is, it makes sense. For example with addresses, you create a new adress or select one first, then you go to the right and maybe copy the address or whatever. With aggregate TXes you make the TX first and then do stuff on the right (edit, delete, send). So the workflow is from left to right, which makes sense. The size of the boxes shouldn't matter, but the functionality.

PR #1051, opened 2021-02-05
Update namespace list pagination to infinite scroll pagination

james-lin71:


rg911 (approved):
@bassemmagdy @yilmazbahadir could you review it?

PR #1059, opened 2021-02-05
Reworked create account modal dialog

myxmaster:
Instead of opening an additional modal dialog for entering the password the password input is now included in the create account modal dialog.

While creation is in progress, a loading animation is shown on the confirm button and all inputs are disabled.

grafik


rg911:
Nice work

rg911 (approved):

PR #1060, opened 2021-02-05
Improved backup dialog layout and translations

myxmaster:
Show a loading animation while backup file is created, improved layout and English translation.

Japanese and Chinese translations should be updated as well.

grafik


rg911 (approved):
👍

PR #1069, opened 2021-02-07
Removed unnecessary notifications

myxmaster:
There were lots of unnecessary success notifications during profile creation and import. In case of an error a notification makes sense, but not if everything is ok.

grafik

grafik


myxmaster:
When creating a ledger profile, there is also such a notification:

FormProfileCreation line 228:

    this.$store.dispatch('notification/ADD_SUCCESS', NotificationType.OPERATION_SUCCESS);

I can't test this so I wasn't sure if it's ok to delete this one also?

rg911 (approved):

PR #1074, opened 2021-02-08
Remove unnecessary notification

myxmaster:
Follow-up for #1069


rg911 (commented):
@bassemmagdy @yilmazbahadir @TriPhan96 

Is it safe to remove this notification? Ledger is the main concern

TriPhan96:
I agree with this change.
But there is one issue in Ledger profile creation flow, there is no `Terms and Conditions & Privacy policy` option. This option should be added in Ledger profile creation flow

rg911 (approved):

PR #1077, opened 2021-02-08
auto sync to working nodes

bassemmagdy:


src/store/Network.ts
```diff
@@ -255,15 +255,18 @@ export default {
             const currentProfile: ProfileModel = rootGetters['profile/currentProfile'];
             const networkService = new NetworkService();
             const nodeService = new NodeService();
+            if (!newCandidate) {
```
rg911, line 258: This will not fix the root issue, right? no matter it is new or not it should retry ->. timeout -> switch. 

also instead of randomly pick, it would be better to build a random node list to try 1 by 1 after certain seconds elapse. So the user get to connect to the best node

src/store/Network.ts
```diff
@@ -255,15 +255,18 @@ export default {
...
+                newCandidate = networkConfig.nodes[Math.floor(Math.random() * networkConfig.nodes.length)].url;
+            }
             const networkModelResult = await networkService
                 .getNetworkModel(newCandidate, (currentProfile && currentProfile.generationHash) || undefined)
                 .toPromise();
             if (!networkModelResult) {
-                throw new Error('Connect error, active peer cannot be found');
+                await dispatch('CONNECT');
```
yilmazbahadir, line 265: This might cause an infinite loop during a network connection failure?

rg911, line 265: there should be a throw I guess, e.g. user's own network problem. otherwise it is going to try indefinitely. 
there should be a final cut off and throw

rg911 (changes requested):

src/store/Network.ts
```diff
@@ -251,20 +253,56 @@ export default {
...
-            if (!networkModelResult) {
-                throw new Error('Connect error, active peer cannot be found');
-            }
-            const { networkModel, repositoryFactory, fallback } = networkModelResult;
-            if (fallback) {
-                throw new Error('Connection Error.');
+            let nodeNetworkModelResult: any;
+            if (newCandidate) {
+                nodeNetworkModelResult = await networkService
+                    .getNetworkModel(newCandidate, (currentProfile && currentProfile.generationHash) || undefined)
```
yilmazbahadir, line 262: If we don't change the current implementation of `networkService.getNetworkModel` it'll find some other node which is available but we don't want this in this block. This block is for 'I have a custom node let's try that and if doesn't work give me an error' .

yilmazbahadir (approved):

rg911 (approved):

PR #1109, opened 2021-02-12
bugfixes

### LINKED ISSUE #1108: Namespace/mosaic transactions doesn't appear in transaction list because of missing rental fees
users have to navigate to namespace/mosaic page first to fetch Estimated rental fees before being able to see their transactions in list

bassemmagdy:
- resolves #1108


src/views/pages/profiles/LoginPageTs.ts
```diff
@@ -254,6 +254,7 @@ export default class LoginPageTs extends Vue {
             await this.$store.dispatch('account/SET_CURRENT_ACCOUNT', defaultAccount);
             this.$store.dispatch('diagnostic/ADD_DEBUG', 'Profile login successful with currentProfileName: ' + currentProfileName);
             $eventBus.$emit('onLogin', currentProfileName);
+            await this.$store.dispatch('network/REST_NETWORK_RENTAL_FEES');
```
rg911, line 257: Does it need to be at the login stage at all? Should just need it when namespace / mosaic transaction right? can it be loaded when ppls announce the transactions?

rg911 (changes requested):

src/views/pages/profiles/LoginPageTs.ts
bassemmagdy: the problem is to show transaction details of an already created mosaic or namespace transaction in dashboard we need to fetch rental fees.

rg911: okay, so if it is loaded from login, should the dispatched be removed from other places?

bassemmagdy: IMO no, because we should always keep track of latest estimations from REST while users are trying to initiate new transactions.

yilmazbahadir: Couldn't get the relation between login and rental fees. If we need the up to date rental fees on tx detail view then why not move there? We should stop adding api calls on login or on LOAD_ACCOUNT IMO, because usually, it's too eager and it leads too much calls in short amount of time(stress at the backend), causing long login wait times for the user.

bassemmagdy: @yilmazbahadir transactions history is fetched on login so, once a user gets redirected to the dashboard/home page if history includes any namespace/mosaic transactions in order to display them on that list rental fees need to be fetched already otherwise, those type of transactions(mosaic-namespace) are not listed and the missing rental fees console error happens. 

Screen Shot 2021-02-15 at 12 20 41 PM

yilmazbahadir: Oh, I thought it as "tx detail page", it is the tx history list, okay in that case. We shouldn't be getting all the tx list on login in the first place IMO but it's part of a big refactoring so fine for now.

rg911: Approve it for now, but we will need to review when refactoring the API call for the wallet as whole

rg911 (approved):

PR #1121, opened 2021-02-14
Allow entering localized amount + display localized amounts

myxmaster:
- AmountDisplay now always displays amounts with at least one decimal and without trailing zeroes. Reason: Demonstrate which decimal separator is used to avoid input errors.
- Display fees in harvesting page localized and always with fractional part (only exception: 0 is shown without fractional part in statistics panel)
- Display fees in fee selector localized
- Allow entering localized amount (in transfer/invoice/simple aggregate tx) and validate no thousand's separator is entered to avoid input errors (also created unit tests for new validator). Also tested exporting/importing QR codes with fractional parts and thousand's separators with English and German localization.
- Fixed fractional part not being displayed when editing simple aggregate tx

Screenshots with German localization:

grafik

grafik

grafik

grafik


rg911 (commented):
need to resolve conflicts

rg911 (approved):

PR #1122, opened 2021-02-14
Use toasts for notifications

### LINKED ISSUE #976: Notification popups can be missed too easily
There are some issues with the notifications that are displayed in the top right corner:
- The timeout is too short. At least for errors there should either be no timeout or you should be able to look it up in some place.
- New notifications replace the previously shown notification. So some notifications might be displayed for a very short time so that you are not able to read it.

I suggest using a library like this:
https://maronato.github.io/vue-toastification/

myxmaster:
Now using toasts from package vue-toastification for notifications instead of notice from iview.

- Notifications are now stackable. Previously only the newest one was visible.
- Error notifications now don't have a timeout and therefore cannot be missed. They have to be clicked to go away. All other notifications have a timeout of 3 seconds which is paused when the toast is hovered.

This fixes #976.

grafik


myxmaster:
I created an issue for those redundant error messages: #1124.

rg911 (commented):
Can we check if this will work with Ledger pls? @TriPhan96 @cryptoBeliever

cryptoBeliever:
I think before merging we should fix https://github.com/nemgrouplimited/symbol-desktop-wallet/issues/1124

myxmaster:
I don't think we need to fix #1124 before merging this, because this just makes these problems visible. It's not better to just hide duplicate messages. But I agree that #1124 should get a high priority.

myxmaster:
Fixed type check errors using solution from here: https://github.com/Maronato/vue-toastification/issues/67
Also had to disable some lint rules for one file.

rg911 (approved):

PR #1138, opened 2021-02-16
Node connection retry, create profile & login when offline

yilmazbahadir:
**Node (re)connection flow :**
1. At the first login, we randomly pick a node from the static list(we have a list for each network type) and try to connect, if that one fails we pick another one until we find an available one.
2. If we find one, set it as selectedNode for the profile, at the next login we’ll first use the same one to connect and then fallback to the node list(random pick). If the user selects a different node from the nodeSelector after the login, we’ll set this node as selectedNode.
3. During the node connection failure and trying on different nodes, we show the progress to the user.
4. If none of the nodes was available we show a warning message with an option to re-try.

for a quick demo: https://www.loom.com/share/da590e484d8a443095a42da80fdbadeb

**Changed**:
* there is no more “default node” definition in the wallet, the wallet shouldn’t query nodes before the user logs in. The only exception to this is the “profile import>select accounts” step where it queries the node to get the account balances. 
* Multi-network support (Testnet and Mainnet configurations )
* Removed MIJIN networks from the list (as DH and DM suggested)
* Removed top menu from nonprotected pages(profile create/import vs)
* Profile create/import and login when offline


src/views/layout/RouterPage.vue
```diff
@@ -1,6 +1,23 @@
...
+                </div>
+                <transition name="fade" mode="out-in">
+                    <div class="main-outer-container">
+                        <div class="page-wrapper">
+                            <router-view />
+                        </div>
+                    </div>
+                </transition>
+                <div class="footer">
+                    <span class="footer-phrase">Powered by <b>NEM Technology.</b></span>
```
rg911, line 17: copyright?

Also NEM Technology doesn't seem right to me, NEM Software perhaps

src/store/AppInfo.ts
```diff
@@ -50,8 +51,8 @@ const appInfoState: AppInfoState = {
     loadingDisableCloseButton: false,
     hasControlsDisabled: false,
     controlsDisabledMessage: '',
-    faucetUrl: networkConfig.faucetUrl,
-    settings: settingService.getProfileSettings(ANON_PROFILE_NAME),
+    faucetUrl: networkConfig[NetworkType.TEST_NET].faucetUrl,
```
rg911, line 54: why hardcoded here? should read from current profile perhaps?

rg911 (changes requested):
Looks good 👍

src/views/layout/RouterPage.vue
yilmazbahadir: RouterPage.vue is basically copy of PageLayout.vue as unprotected(without menu), that's why we have it here too.
image

src/store/AppInfo.ts
yilmazbahadir: Since the Faucet won't be available in mainnet, I didn't work on tying it to the profile network type. I can work on it if it doesn't make sense.

src/views/layout/RouterPage.vue
rg911: Hmm, I think that's wrong, Should be changed to NEM Software, we don't have a NEM Technology entity ?

src/store/AppInfo.ts
rg911: Is it a big change to use profile network type? In all I think it doesn't makes sense to have networktype hardcoded anymore

src/views/layout/RouterPage.vue
yilmazbahadir: fixed => Copyright © 2021 NEM Software

src/store/AppInfo.ts
yilmazbahadir: fixed

rg911 (approved):

PR #1179, opened 2021-02-19
fix bugs

### LINKED ISSUE #1140: "Delete Contact" dialog appears again when a user clicks on a contact after removing another one
**Steps to reproduce:**

0. You may import the address book if no contacts in a profile
1. Delete any of the contacts
2. Click on the another one remaining in the address book

**Actual result:** The Delete Contact dialog appears again until dismissed
Note: this issue occurs also after deleting all contacts and importing an address book again, then clicking on any of the entries
[Video](https://share.getcloudapp.com/yAu66p2P)

### LINKED ISSUE #1162: Hints / messages that are displayed in the upper right corner of the wallet are not meaningful enough.
I think it should generally be checked whether the hints / messages that are displayed when using the wallet are really meaningful enough. The messages should be adapted for each case.

Here is an example:

Under "Profile Import" there is no direct input field. Instead of  "Fields cannot be empty. Please, verify your input." 
a message should be displayed that describes that the user has to select an address.

---
profile_import2

### LINKED ISSUE #1170: No default network when creating profile
I think for now we should keep Testnet as a default network when creating a profile.
For mainnet launch release we should change to Mainnet.

For now when I'm opening live build in incognito default is empty.

image

### LINKED ISSUE #1178: Can't scan QR codes on 0.14.0 Mac DMG
Screen Shot 2021-02-19 at 20 04 05

Can't open webcam to scan QR code on MacOS release files. Work on development version (web (npm run dev) and electron (npm run start)

bassemmagdy:
- fixes #1178 
- fixes #1162 
- fixes #1170 
- fixes #1140

### DIFF
--- a/src/language/en-US.json
+++ b/src/language/en-US.json
@@ -840,7 +840,7 @@
     "seed_account_can_not_be_more_than_10": "You have reached the limit of accounts that can be created in one profile. Please use a different profile to create a new account.",
     "select_a_namespace": "Select a namespace",
     "select_a_profile": "Select a profile",
-    "select_accounts": "Select Accounts",
+    "select_accounts": "Selected Accounts",
     "select_all": "Check all",
     "select_contact": "Select contact",
     "send": "Send",
@@ -1037,5 +1037,6 @@
     "click_to_retry": "click to retry",
     "current_node_is_not_available": "Current node is not available. Trying to connect to node #{0}/{1}",
     "nodes_unavailable_check_network_and": "Nodes({0}/{1}) are currently unavailable. Please check your network connection, settings and ",
-    "copyright": "Copyright © 2021 NEM Software"
+    "copyright": "Copyright © 2021 NEM Software",
+    "address_list_cannot_be_empty": "Selected accounts cannot be empty. Please, select 1 or more accounts from the list."
 }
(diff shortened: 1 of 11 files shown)


src/language/en-US.json
```diff
@@ -1037,5 +1037,6 @@
     "click_to_retry": "click to retry",
     "current_node_is_not_available": "Current node is not available. Trying to connect to node #{0}/{1}",
     "nodes_unavailable_check_network_and": "Nodes({0}/{1}) are currently unavailable. Please check your network connection, settings and ",
-    "copyright": "Copyright © 2021 NEM Software"
+    "copyright": "Copyright © 2021 NEM Software",
+    "address_list_cannot_be_empty": "Selected accounts cannot be empty. Please, select 1 or more accounts from the list."
```
segfaultxavi, line 1041:
```suggestion
    "address_list_cannot_be_empty": "Please, select 1 or more accounts from the list."
```

I think this sentence on its own is enough. Otherwise, it sounds like "the accounts are empty", as in, they have no funds.

segfaultxavi (changes requested):

rg911 (dismissed):

rg911 (approved):

PR #1192, opened 2021-02-21
Changed dashboard icons

myxmaster:
Changed and harmonized icons (incl. colors) in table with those in navigation tabs

grafik


segfaultxavi:
Aren't the icons reversed?
Shouldn't **Send** be the **Outgoing** icon (down arrow) and **Invoice** the **Incoming** (Up arrow)?

myxmaster:
hahah yes of course, will fix later.

rg911 (approved):

jabo38:
I really like the tabs saying "History/Send/Invoice" as those are more accurate terms, but I'm personally not a fan of the red and green. I think it clashes with the purple theme.

myxmaster:
Ok, I understand that. The intention was to make it as clear as possible.

Anyway, we should talk about 2 things still:
1) The history icon is barely recognizable on PC screens (with normal resolution). Even in the mobile wallets you can't see those little dots, because it is just too small:
grafik
2) Shouldn't the icons for incoming and outoing transactions (in the transaction table) be the same as for "Send" and "Invoice" above the table (navigation tabs)?

Generally the icons should be the same for both, mobile and desktop wallet. So if we can agree to any changes, we should do it for both always.

jabo38:
I agree that the icons should be the same across the different apps to give users a communing feeling. 

The Send and Receive icons in the mobile app are great to me. 

The history icon is definitely lack.  If I showed that icon to many people, I'm not sure any would guess history. 

This icon looks more like a list of txs to me than the one currently used. 

image

myxmaster:
Yes, that is a lot better!

PR #1201, opened 2021-02-22
Login page improvements

myxmaster:
- Fixed error message that is shown when entered profile name does not exist (Chinese is verified, Japanese not)
- Changed inputs background to gray (as elsewhere in the application)

grafik

### DIFF
--- a/src/views/pages/profiles/LoginPageTs.ts
+++ b/src/views/pages/profiles/LoginPageTs.ts
@@ -37,6 +37,7 @@ import { SettingService } from '@/services/SettingService';
 import { SettingsModel } from '@/core/database/entities/SettingsModel';
 import { AccountService } from '@/services/AccountService';
 import { NetworkTypeHelper } from '@/core/utils/NetworkTypeHelper';
+import _ from 'lodash';
 
 @Component({
     computed: {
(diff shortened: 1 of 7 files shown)


src/views/pages/profiles/LoginPageTs.ts
```diff
@@ -118,21 +119,17 @@ export default class LoginPageTs extends Vue {
...
-
-        this.profilesClassifiedByNetworkType = this.profiles.reduce(reducer, []);
         if (!this.profiles.length) {
             return;
         }
-        // accounts available, iterate to first profiles
+
+        const profilesGroupedByNetworkType = _.groupBy(this.profiles, (p) => p.networkType);
+        this.profilesClassifiedByNetworkType = Object.values(profilesGroupedByNetworkType).map((profiles) => ({
+            networkType: profiles[0].networkType,
```
rg911, line 128: only the top one?

myxmaster: Since `profilesGroupedByNetworkType` is grouped by network type, all the profiles have the same network type... 😀 
So yes, the network type can be retrieved from the first profile.

rg911 (approved):

PR #1209, opened 2021-02-24
Namespace duration fix

### LINKED ISSUE #1208: Incorrect Min Namespace Duration shown
In the Namespace creation form, an invalid Min Namespace Duration is shown:
image
It says 172800 (blocks, I assume), but at 30s per block that is 60 days, not 30 as the network announces:

http://api-01.ap-northeast-1.testnet.symboldev.network:3000/network/properties
```
"blockGenerationTargetTime":"30s"
"minNamespaceDuration":"30d"
```

This is already confusing users, which requested changes in the docs to match what the Wallet says: https://github.com/nemtech/symbol-docs/issues/584

bassemmagdy:
- fixes #1208


src/config/NetworkConfig.ts
```diff
@@ -93,10 +93,10 @@ const defaultMainnetNetworkConfig: NetworkConfig = {
         namespaceGracePeriodDuration: 2592000,
         lockedFundsPerAggregate: '10000000',
         maxCosignatoriesPerAccount: 25,
-        blockGenerationTargetTime: 15,
+        blockGenerationTargetTime: 30,
         maxNamespaceDepth: 3,
         maxMosaicDuration: 21024000,
-        minNamespaceDuration: 172800,
+        minNamespaceDuration: 86400,
         maxNamespaceDuration: 2102400,
```
rg911, line 100: "minNamespaceDuration": "30d",
"maxNamespaceDuration": "1825d",

Max value doesn't seem right to me

Also, NamespaceDuration is network properties, would it be better to calculate from network/properties?

rg911 (changes requested):

src/views/forms/FormNamespaceRegistrationTransaction/FormNamespaceRegistrationTransactionTs.ts
```diff
@@ -144,7 +144,7 @@ export class FormNamespaceRegistrationTransactionTs extends FormTransactionBase
         this.formItems.registrationType = this.registrationType || NamespaceRegistrationType.RootNamespace;
         this.formItems.newNamespaceName = this.namespaceId ? this.namespaceId.fullName : '';
         this.formItems.parentNamespaceName = this.parentNamespaceId ? this.parentNamespaceId.fullName : '';
-        this.formItems.duration = this.duration || 172800;
+        this.formItems.duration = this.duration || 86400;
```
yilmazbahadir, line 147: To access the connected node's network properties we can use `this.networkConfiguration`.

`this.formItems.duration = this.duration || this.networkConfiguration.minNamespaceDuration || 86400` ?

src/config/NetworkConfig.ts
yilmazbahadir: I believe these are the default values and is getting overwritten by the connected node's network properties.

yilmazbahadir (approved):

rg911 (approved):

PR #1210, opened 2021-02-24
Message with 1024 characters not possible

james-lin71:


src/core/validation/validators/MaxMessageValidator.ts
```diff
@@ -0,0 +1,31 @@
...
+    /**
+     * Validates the max bytes of message
+     * @static
+     * @param {*} value
+     * @param {number} maxMessageNumber
+     * @returns {boolean}
+     */
+    public static validate(value: any, maxMessageNumber: number): boolean {
+        const bytes: number = new TextEncoder().encode(value).length;
+        return bytes < maxMessageNumber;
```
rg911, line 29: this is not quite right,  there is 1 byte (message type) appended to the message, so it will be 1 byte less

rg911 (changes requested):

bassemmagdy (approved):

rg911 (approved):

PR #1211, opened 2021-02-24
fix alias incoming transaction direction

### LINKED ISSUE #784: Incoming transaction send to alias presented as outgoind on receiver side
Steps:
1. Send the transaction from account A to namespace alias (from outside wallet because right now it's not possible) linked with account B.
2. Switch to account B and go to transactions.

Incoming transaction is presented as outgoing. 

image
image

We need to resolve namespace at time of sending. Not sure if API allows that. To fix this small issue lot of work may be required, I'm reporting anyway to have this in queue.

bassemmagdy:
- fix #784

### DIFF
--- a/src/core/transactions/ViewTransferTransaction.ts
+++ b/src/core/transactions/ViewTransferTransaction.ts
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and limitations under the License.
  *
  */
-import { PublicAccount, TransferTransaction } from 'symbol-sdk';
+import { NamespaceId, PublicAccount, TransactionType, TransferTransaction } from 'symbol-sdk';
 // internal dependencies
 import { TransactionView } from './TransactionView';
 import { AttachedMosaic } from '@/services/MosaicService';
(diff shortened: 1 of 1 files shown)


src/core/transactions/ViewTransferTransaction.ts
```diff
@@ -25,9 +25,23 @@ import { MosaicModel } from '@/core/database/entities/MosaicModel';
 export class ViewTransferTransaction extends TransactionView<TransferTransaction> {
     public get isIncoming() {
         const currentSignerAddress = this.$store.getters['account/currentSignerAddress'];
+        if (this.transaction.type === TransactionType.TRANSFER && this.transaction.recipientAddress instanceof NamespaceId) {
+            const linkedAddress = this.$store.getters['namespace/linkedAddress'];
+            if (!linkedAddress) {
+                this.checkLinkedAddress(this.transaction.recipientAddress);
+                return false;
```
rg911, line 32: why return false here, does it needs to recheck the linkedAddress after dispatch called?

bassemmagdy: yes, as GET_LINKED_ADDRESS is an async function and doesn't return the linked address so there must be a return statement to avoid sending multiple requests to fetch namespace linkedAdress.

rg911 (approved):

PR #1215, opened 2021-02-25
Fix ci build

bassemmagdy:


rg911 (commented):
does this fixing the CI build on main?

src/components/AmountDisplay/AmountDisplayTs.ts
```diff
@@ -42,7 +42,11 @@ export class AmountDisplayTs extends Vue {
 
     /// region computed properties getter/setter
     get integerPart(): string {
-        return this.value >= 0 ? Math.floor(this.value).toLocaleString() : Math.ceil(this.value).toLocaleString();
+        return this.value >= 0
```
yilmazbahadir, line 45: Would this do the job?
```
return this.value >= 0
            ? Math.floor(this.value).toLocaleString() : '-' + Math.floor(this.value * -1).toLocaleString()
```

bassemmagdy:
> does this fixing the CI build on main?

that should fix unit tests for ci-integration

src/components/AmountDisplay/AmountDisplayTs.ts
bassemmagdy: yes, the same thing but looks better!

yilmazbahadir (approved):

PR #1225, opened 2021-02-26
Fetch harvesting status periodically 

### LINKED ISSUE #1222: Show accurate Harvesting status by periodic polling
It seems like whether a node has accepted an account's persistent delegation requests or not is only updated when entering the Harvesting screen.
This means that the user needs to go back and forth between Home and Harvesting for a while until the node updates, which might never happen.

It would be easier if the node was periodically polled (say every 30s) to reflect the current state of the link.
This would only be needed in the "ACTIVATION IN PROGRESS" and "ACTIVE" states.

bassemmagdy:
fixes #1222

### DIFF
--- a/src/components/HarvestStatisticsPanel/HarvestStatisticsPanelTs.ts
+++ b/src/components/HarvestStatisticsPanel/HarvestStatisticsPanelTs.ts
@@ -40,12 +40,12 @@ export class HarvestStatisticsPanelTs extends Vue {
     private isFetchingUnlocked: any;
     private isFetchingBlocks: any;
 
-    async created() {
-        await this.refresh();
+    created() {
+        this.refresh();
     }
 
-    public async refresh() {
-        await this.refreshHarvestingStatus();
+    public refresh() {
+        this.refreshHarvestingStatus();
         this.refreshHarvestingStats();
     }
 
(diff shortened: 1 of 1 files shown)


src/components/HarvestStatisticsPanel/HarvestStatisticsPanelTs.ts
```diff
@@ -37,39 +37,63 @@ export class HarvestStatisticsPanelTs extends Vue {
...
+            this.isFetchingUnlocked = setInterval(async () => {
+                await this.refreshHarvestingStatus();
+            }, 45000);
+        });
+    }
+
+    private refreshStatusBlocks(): void {
+        this.$nextTick(async () => {
+            this.isFetchingBlocks = setInterval(async () => {
+                await this.refresh();
```
yilmazbahadir, line 88: should we use `refreshHarvestingStats`instead of `refresh` here?

src/components/HarvestStatisticsPanel/HarvestStatisticsPanelTs.ts
```diff
@@ -37,39 +37,63 @@ export class HarvestStatisticsPanelTs extends Vue {
     public harvestedBlockStats: HarvestedBlockStats;
     public isFetchingHarvestedBlockStats: boolean;
     public networkCurrency: NetworkCurrencyModel;
+    private isFetchingUnlocked: any;
+    private isFetchingBlocks: any;
 
-    created() {
-        this.refresh();
+    async created() {
```
yilmazbahadir, line 43: Should we check if it works without all the async/await pairs in the class?

yilmazbahadir (approved):

PR #1226, opened 2021-02-26
Profile import account selection: show balance zero instead of N/A

myxmaster:
grafik


bassemmagdy:
@myxmaster this solution will still produce the same issue if profile is not connected to any of the nodes.

myxmaster:
If the wallet is not connected to any node, there is no repositoryFactory so there will be an error when calling `repositoryFactory.createAccountRepository()`.

If a connection to a node could be established but is lost before trying to retrieve the account infos, `getAccountsInfo(this.addressesList)` throws an error and N/A is shown,

rg911 (approved):

PR #1229, opened 2021-02-27
Move logout button from settings to top navbar

myxmaster:
grafik


myxmaster:
Closing because combining this with addition of profile deletion button.

PR #1235, opened 2021-03-01
Disable password copy

### LINKED ISSUE #1167: Copy & Past of the password when creating the wallet should not be accepted.
v0.14.0 linux

This should not be possible.

bassemmagdy:
- disable copy, paste, cut and select in password fields -> fixes #1167

### DIFF
--- a/src/components/QRCode/QRCodePassword/QRCodePassword.vue
+++ b/src/components/QRCode/QRCodePassword/QRCodePassword.vue
@@ -22,6 +22,10 @@
                                         <input
                                             v-model="formItems.password"
                                             v-focus
+                                            onselectstart="return false"
+                                            onpaste="return false;"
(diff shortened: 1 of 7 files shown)


src/components/QRCode/QRCodePassword/QRCodePassword.vue
```diff
@@ -22,6 +22,10 @@
                                         <input
                                             v-model="formItems.password"
                                             v-focus
+                                            onselectstart="return false"
```
yilmazbahadir, line 25: Should we create a new component `PasswordInput` and have these all common properties in the new component with defaults? And use that 'PasswordInput' component everywhere?

bassemmagdy:
fixed in https://github.com/nemgrouplimited/symbol-desktop-wallet/pull/1372

PR #1250, opened 2021-03-04
Do not reset transfer tx form on signer change

myxmaster:
Reset was added in #55 but it is unclear why. It works without reset and this is more convenient (as noticed by @jabo38).

Also fixed validation error messages.


rg911 (commented):
@bassemmagdy @cryptoBeliever Do we know why it reset the transaction form on changing signer?

cryptoBeliever:
@rg911 
If comment "Is necessary to make the mosaic inputs reactive" is not valid then I don't know :)

myxmaster:
The mosaic input works fine, I tested that. So yeah, no idea what that comment was for...

PR #1255, opened 2021-03-06
Update support Ledger Symbol app to version 0.0.9

TriPhan96:
- Add Ledger profile type at Symbol profile creation step
- Add Ed25519 curve option when derive Ledger account
- Update Ledger app version checking function


src/core/utils/Ledger.ts
```diff
@@ -180,19 +204,21 @@ export class SymbolLedger {
...
      * @returns respond package from Ledger
      */
-    private async ledgerMessageHandler(path: string, rawTx: Buffer, chainCode: boolean) {
+    private async ledgerMessageHandler(path: string, rawTx: Buffer, chainCode: boolean, profileType: number) {
         const TX_INS_FIELD = 0x04;
         const MAX_CHUNK_SIZE = 255;
         const CONTINUE_SENDING = '0x9000';
 
-        const curveMask = 0x80; // use Curve25519
+        const curveMask = profileType == LedgerProfileType.NORMAL ? 0x80 : 0x40;
```
rg911, line 221: not sure if this is correct but I might be wrong. 
the `0x40` is the legacy BTC curve right?
so publicKey generation should always be the same as before as it is correct, the only different bit of the new `Normal` type should be the private key generation off the mnemonic?

TriPhan96: These constants are just our definitions, `0x40` is for the Secp256K1 curve and `0x80` is for the Ed25519 curve. Ref [here](https://github.com/nemgrouplimited/symbol-ledger-app/blob/213b53908a0b659cdac9d0e3a3a917b930cde8d1/src/apdu/constants.h#L31-L32)

The publicKeys generate from `0x40` constant will be the same as before. And the correct Ed25519 curve is used when use `Normal` type account (`0x80` constant)

src/language/en-US.json
```diff
@@ -729,6 +732,7 @@
     "option_privatekey_account": "I want to import an existing account private key",
     "option_root_namespace": "Root Namespace",
     "option_sub_namespace": "Sub-namespace",
+    "optin_wallet": "Optin wallet",
```
segfaultxavi, line 735:
```suggestion
    "optin_wallet": "Opt-in wallet",
```

PR #1266, opened 2021-03-08
Re subscribe ws

bassemmagdy:


src/services/RESTService.ts
```diff
@@ -47,8 +48,14 @@ export class RESTService {
         // open websocket connection
         const address = Address.createFromRawAddress(addressStr);
         if (listener) {
-            if (listener && !listener.isOpen()) {
-                await listener.open();
+            if (!listener.isOpen()) {
+                await listener.open(async (event: { client: string; code: any; reason: any }) => {
+                    if (event.code !== 1005) {
```
rg911, line 53: should check if event is not undefined as well?

rg911 (approved):

PR #1268, opened 2021-03-08
fix mnemonic import bug

bassemmagdy:


rg911:
Go with 1269, but thanks very much mate 👍

PR #1269, opened 2021-03-08
Fix Import mnemonic bad accounts

Peersyst:


src/views/pages/profiles/import-profile/account-selection/AccountSelectionTs.ts
```diff
@@ -251,6 +219,10 @@ export default class AccountSelectionTs extends Vue {
         const key = this.currentProfile.networkType === NetworkType.MAIN_NET ? 'mainnet' : 'testnet';
         const whitelisted = process.env.KEYS_WHITELIST[key];
         const optInAccounts = possibleOptInAccounts.filter((account) => whitelisted.indexOf(account.publicKey) >= 0);
+        console.log(this.currentMnemonic);
```
rg911, line 222: shall we remove the console log?

src/views/pages/profiles/import-profile/account-selection/AccountSelectionTs.ts
```diff
@@ -175,39 +175,7 @@ export default class AccountSelectionTs extends Vue {
         if (!this.selectedAccounts.length) {
```
bassemmagdy, line 175: this condition needs to check optIn accounts too

Peersyst: I don't think so, we want to be sure that at least there's one seed account with the correct curve imported

Peersyst: But user may not want to import opt in account (either if doesn't have one or if he moved the funds to another)

src/views/pages/profiles/import-profile/account-selection/AccountSelectionTs.ts
Peersyst: Sure! My bad

rg911 (dismissed):

src/views/pages/profiles/import-profile/account-selection/AccountSelectionTs.ts
rg911: @bassemmagdy I think @AdriaCarrera is right, as the opt in keys are loaded as env vars, it will always show.

rg911 (approved):

PR #1270, opened 2021-03-09
Fix ws bug

bassemmagdy:


rg911 (changes requested):
Not sure this will work, so
- if rest is down, obviously the listener won't connect, but currently in wallet, user won't get notified and can still announce transactions. 
- there should be an error handler on transaction announce, if errors trapped (rest call), then notify the user node is not connectable. and probably change the node status to red circle. 
- so user can choose different node to try.

PR #1273, opened 2021-03-09
Update support Ledger Symbol app to version 0.0.9

TriPhan96:
- Add Ed25519 curve option when derive Ledger account
- Update Ledger app version checking function


rg911:
will open a new PR from upstream branch

PR #1315, opened 2021-03-14
Remove add private key account option for Ledger profile

TriPhan96:


rg911:
closing it as another PR from @bassemmagdy  did the same thing

PR #1395, opened 2021-04-05
Resolved #1389 simple offline transaction not possible

### LINKED ISSUE #1389: Offline transaction - Simple transaction not possible offline
Wallet version: 1.0.1 for Linux

Steps:

1. Clear all profiles/settings (Linux ~/.symbol-desktop-wallet)
2. Open wallet
3. Create wallet
4. Logout
5. Go to Offline Transaction (Simple Transaction tab) - Offline Transfer transaction view is not visible entirely

image

6. Switch to Cosign transaction view.
7. Switch back to "Simple transaction view". You got information "The wallet could not load data about the network currency (e.g. symbol:xym), please connect to a valid node.". And not possible to init transaction.

image

Peersyst:


src/services/MosaicService.ts
```diff
@@ -226,15 +226,15 @@ export class MosaicService {
...
+        // const storedNetworkCurrencies = this.networkCurrencyStorage.get(generationHash);
         return repositoryFactory.getCurrencies().pipe(
             map((networkMosaics) => {
                 const currency = this.getNetworkCurrency(networkMosaics.currency);
                 const harvest = this.getNetworkCurrency(networkMosaics.harvest);
                 return new NetworkCurrenciesModel(currency, harvest);
             }),
             tap((d) => this.networkCurrencyStorage.set(generationHash, d)),
-            ObservableHelpers.defaultFirst(storedNetworkCurrencies),
+            // ObservableHelpers.defaultFirst(storedNetworkCurrencies),
```
rg911, line 237: remove comments?

src/services/offline/MockModels.ts
```diff
@@ -147,3 +150,15 @@ export const OfflineAccountNames = (address: Address) => new AccountNames(addres
 export const OfflineNamespaceNames = (namespaceId: NamespaceId) => new NamespaceName(namespaceId, 'mocknamespace');
```
rg911, line 150: OfflineNamespaceNames why this is using mocknames?

Peersyst: There's a request done for some NamespaceIds when selecting a profile/account, so if you don't return an array with the same size of the ids request it breaks.

src/services/MosaicService.ts
Peersyst: Yes, thanks

rg911 (changes requested):
have you checked the offlineNetworkProperties against the new testnet?  There seems to be some misalignments in between

src/services/offline/MockModels.ts
```diff
@@ -147,3 +150,15 @@ export const OfflineAccountNames = (address: Address) => new AccountNames(addres
...
 
 export const OfflineMultisigAccountGraphInfo = new MultisigAccountGraphInfo(new Map());
+
+export const OfflineNetworkCurrencies = (networkType: NetworkType): NetworkCurrencies => {
+    const publicCurrency = new Currency({
```
fboucquez, line 155: These values should be loaded from the NetworkConfig.ts instead of having a copy. 
Also, remember we would like to support private networks eventually.

rg911 (approved):

PR #1401, opened 2021-04-06
New Testnet and Mainnet network config fix

fboucquez:
- Updated testnet and mainnet configuration in https://github.com/nemgrouplimited/symbol-desktop-wallet/blob/main/src/config/NetworkConfig.ts
- Improved storage so it can be unit tested.
- Migrated testnet profile storage to new testnet
- Removed old tesnet nodes from node storage.
- Added unit testing around Profile and Node storage migration.
- Removed hardcoded testnet peer list.

There are also hardcoded values in https://github.com/nemgrouplimited/symbol-desktop-wallet/blob/main/src/services/offline/MockModels.ts#L28 that are revisited in https://github.com/nemgrouplimited/symbol-desktop-wallet/pull/1395

Redo of https://github.com/nemgrouplimited/symbol-desktop-wallet/pull/1390


fboucquez:
If we need to keep the static list, we could put it in the NetworkConfig for both mainnet and testnet, Network.ts can do the merging.

rg911 (approved):

PR #1405, opened 2021-04-07
Revert "New testnet Configuration"

bassemmagdy:
Reverts nemgrouplimited/symbol-desktop-wallet#1390


rg911 (approved):

fboucquez (changes requested):
There is no need to revert this. The node list is not stored in the profile, it's in the node localstorage which has other issues

PR #1411, opened 2021-04-08
Harvesting main fixes

### LINKED ISSUE #1370: Harvesting main fix
### Harvesting fixes:

- [x] **Differentiation between delegate harvesting and remote harvesting:** Users need to understand the differences between delegated harvesting and remote harvesting. Right now both clients does not explain the difference and they merge both types of harvesting into the same screens. We need to separate the key links from the main delegated harvesting screen, and alert the user when clicking to key links saying that they are entering an "expert area"

- [x] **Persistent Delegation Request Timeout**: When a persistent delegation request is sent, we need to check that it is accepted. To do so, we are going to query the node unlockedaccounts every X minutes and check if it has been accepted. If it has not been accepted in 10-15 minutes we will consider that the persistent delegation request has failed.

See also #1292

- [x] **Portability between desktop and mobile clients**: If the delegated harvesting is started in one device and followed in another client, we must maintain consistency. To do so we have three available solutions:

1. Detect when the user started harvesting on another device (when private keys are not stored in the actual one). Query /node/info to all the nodes in our list to check if there's a public key matching the node linked public key. If we find one then check the /unlockedaccounts entrypoint if we are within the list. If we are in the list the user is harvesting properly, otherwise we say harvesting has failed or has been rejected. If we don't find a node then leave a message saying "Delegated harvesting was started in another device and we are unable to maintain status in the actual one"

2. Detect when the user started harvesting on another device (when private keys are not stored in the actual one). If it has been activated in another device leave a message saying that the harvesting screen is only visible in the device that it has been started. And only leave the option for the user to restart all the harvesting process

3. Adding a text banner: if harvesting activated from other devices, please provide the harvesting node details.
Then we can check the unlocked account when possible (dual node), otherwise tell the user: fetching harvesting status… if it has been activated, please do now reactivate again

See also #1243
See also #1067 

- [x] **Fees and fund waste**: We must allow one option for the users to pick the fees for the transactions made by harvesting page. We must also warn the users that harvesting requests may be rejected by the nodes so they may be wasting funds related with network fees when activating / deactivating delegated harvesting.

See also #1338 

- [x] **Custom node selection and custom keys entering**: We must allow the users to enter custom keys (vrf and remote). Because they might already have linked those keys using another client. Also the option of allowing custom node must be present since all nodes aren't returned by the REST server. Add explorer url

See also #1116
See also #1115 
See also #1358 
See also #1347 

- [x] **Unlocked account notifications**: Users need to understand that nodes may kick them if they don't have all the requirements of importance. Adding a notification when a node is selected that says to the user the probability to be inside. (Checking number of unlocked accounts and their importance average)

Peersyst:
Resolves #1370


src/language/en-US.json
```diff
@@ -297,6 +308,7 @@
     "form_label_multisig_operation_type": "Operation type",
     "form_label_namespace_name": "Namespace name",
     "form_label_network_node_url": "Enter a harvesting node URL here (e.g: http://localhost:3000)",
+    "form_label_network_custom_node": "Type your custom harvesting node URL above",
```
OlegMakarenko, line 311:
```suggestion
    "form_label_network_custom_node": "Type your custom harvesting node URL below",
```

image

OlegMakarenko (commented):
image
I don't able to enter text here

OlegMakarenko (changes requested):
locale files should be updated

PR #1428, opened 2021-04-14
fix fees not showing correctly on qrcodes imports

bassemmagdy:


bassemmagdy:
fixed in https://github.com/nemgrouplimited/symbol-desktop-wallet/pull/1518

Issue #1376, opened 2021-03-26
A user can try change duration for symbol.xym mosaic


Eugenity:
**Steps to reproduce:**

1. Go to Mosaics page
2. Go under 'Create new namespaces' tab
3. Specify 'Namespace Name' field with "symbol"
4. Click Send button

**Expected result:** There is an error message pop-up informing a user that such operation for reserved namespaces is imposible
**Actual result:** An operation is initiated but fails with an error message about invalid duration (which literally means that it tries to change the 'symbol' mosaic duration as far as I can assume)

[Video](https://share.getcloudapp.com/YEuRjywb)

cryptoBeliever:
Not fixed. My comments in https://github.com/nemgrouplimited/symbol-desktop-wallet/pull/1452

bassemmagdy:
related to https://github.com/nemtech/catapult-rest/issues/631

cryptoBeliever:
Ok, As a temp fix it's ok :+1:

cryptoBeliever:
Fixed in integration branch.

PR #1452, opened 2021-04-21
fix validation for namespace name #1376

bassemmagdy:


src/language/en-US.json
```diff
@@ -397,7 +397,7 @@
     "Failure_Metadata_Value_Size_Delta_Too_Large": "Validation failed because the metadata value size delta is larger in magnitude than the value size",
     "Failure_Metadata_Value_Size_Delta_Mismatch": "Validation failed because the metadata value size delta does not match expected value based on the current state",
     "Failure_Metadata_Value_Change_Irreversible": "Validation failed because a metadata value change (truncation) is irreversible",
-    "Failure_Mosaic_Invalid_Duration": "Validation failed because the duration has an invalid value",
+    "Failure_Mosaic_Invalid_Duration": "Validation failed because the duration of mosaic has an invalid value",
```
segfaultxavi, line 400:
```suggestion
    "Failure_Mosaic_Invalid_Duration": "Validation failed because the mosaic duration has an invalid value",
```

src/language/en-US.json
```diff
@@ -432,7 +432,7 @@
     "Failure_Multisig_Loop": "Validation failed because a multisig loop is created",
     "Failure_Multisig_Max_Multisig_Depth": "Validation failed because the max multisig depth is exceeded",
     "Failure_Multisig_Operation_Prohibited_By_Account": "Validation failed because an operation is not permitted by a multisig account",
-    "Failure_Namespace_Invalid_Duration": "Validation failed because the duration has an invalid value",
+    "Failure_Namespace_Invalid_Duration": "Validation failed because the duration of namespace has an invalid value",
```
segfaultxavi, line 435:
```suggestion
    "Failure_Namespace_Invalid_Duration": "Validation failed because the namespace duration has an invalid value",
```

segfaultxavi (changes requested):
Is there an explanation somewhere in the UI of **what is a valid duration**?
I mean, do users know what values are valid? If not, telling them "this is invalid" will not help them very much.

src/language/en-US.json
bassemmagdy: @segfaultxavi there is already validation added to the input fields so users will know immediately anyways if they're out of range

src/core/validation/ValidationRuleset.ts
```diff
@@ -45,7 +45,7 @@ export const createValidationRuleSet = ({
         }|maxNamespaceDuration`,
         namespaceName: {
             required: true,
-            regex: '^[a-z0-9]{1}[a-z0-9-_]{0,63}$',
+            regex: '^(?!symbol$)([a-z0-9]{1}[a-z0-9-_]{0,63})$',
```
cryptoBeliever, line 48: - symbol is not only reserved name (check node configuration https://github.com/nemtech/symbol-bootstrap/blob/4db690f095aa1e5ed14642ea1583264c1ac9d530/presets/shared.yml#L25)
- current error "The Namespace name field format is invalid" is not clear for user. 
Why we don't fix error message in rest?

bassemmagdy: @cryptoBeliever that was a temp fix from wallet till we get rest updated, also all other reserved namespaces work as expected and don't need to be validated in client-side.
Screen Shot 2021-06-09 at 3 02 21 PM

cryptoBeliever: @bassemmagdy 
I see. Could you link rest issue?

bassemmagdy: https://github.com/nemtech/catapult-rest/issues/631

cryptoBeliever: Thanks. Maybe good idea is add comment in code that after fixing rest issue this condition can be removed.

PR #1457, opened 2021-04-23
Post-launch Opt-in transaction view

OlegMakarenko:
**1. Added Post-launch Opt-in transaction View:**
image
image

**2. Added Post-launch Opt-in transaction caption/type:**
image

**3. Added Op-in transaction icon:**
image

**4. Fixed the Aggregate Bonded inner transactions view overlaps cosignature form bug:**
image


src/components/TransactionDetails/TransactionOptinPayoutDetails.vue
```diff
@@ -0,0 +1,226 @@
+<template>
```
AdriaCarrera, line 1: You should separate logic between different files (.ts, .vue, .less)

AdriaCarrera (approved):

PR #1458, opened 2021-04-23
fix #873

### LINKED ISSUE #873: Namespaces - expired namespaces visible without "Show expired namespaces" checkbox checked
On namespaces list by default expired namespaces are presented.

image

Expected result: Expired namespace hidden by default and visible only if we check "Show expired namespaces"

bassemmagdy:
fix expired namespace shown by default without using show expiry checkbox


src/services/AssetTableService/NamespaceTableService.ts
```diff
@@ -47,20 +48,35 @@ export class NamespaceTableService extends AssetTableService {
...
-                aliasIdentifier: this.getAliasIdentifier(namespaceModel),
-                metadataList: namespaceModel.metadataList || [],
-            };
-        });
+        const showExpired: boolean = this.showExpired;
+        const currentHeight: number = this.currentHeight;
+        return !showExpired
+            ? namespaces
+                  .filter((ns) => ns.endHeight > currentHeight)
+                  .map((namespaceModel) => {
```
OlegMakarenko, line 56: I would suggest not to duplicate mapper and move it to a separate function

src/services/AssetTableService/NamespaceTableService.ts
```diff
@@ -47,20 +48,35 @@ export class NamespaceTableService extends AssetTableService {
...
+                          hexId: namespaceModel.namespaceIdHex,
+                          name: namespaceModel.name,
+                          expiration: expiration,
+                          expired: expired,
+                          aliasType: this.getAliasType(namespaceModel),
+                          aliasIdentifier: this.getAliasIdentifier(namespaceModel),
+                          metadataList: namespaceModel.metadataList || [],
+                      };
+                  })
+            : namespaces.map((namespaceModel) => {
```
OlegMakarenko, line 68: Same here

rg911 (approved):

OlegMakarenko (approved):

PR #1462, opened 2021-04-27
Post-launch Opt-in transaction view. Cosign and config file changes

OlegMakarenko:
- Don't show Opt-in tx view, if this tx needs current account signature
- Don't show Opt-in tx view. if this tx doesn't have inner Transfer with the current account address as a recipient
- Changed config structure:

`keys-whitelist.json`
```
{
    "preLaunchOptin": {
        "testnet": [],
        "mainnet": []
    },
    "nglFinanceBot": {
        "testnet": [],
        "mainnet": []
    }
}
```

### DIFF
--- a/src/views/modals/ModalTransactionCosignature/ModalTransactionCosignatureTs.ts
+++ b/src/views/modals/ModalTransactionCosignature/ModalTransactionCosignatureTs.ts
@@ -25,6 +25,7 @@ import {
     NetworkType,
     TransactionStatus,
     TransactionType,
+    TransferTransaction
 } from 'symbol-sdk';
 import { mapGetters } from 'vuex';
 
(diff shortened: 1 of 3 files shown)


vue.config.js
```diff
@@ -25,32 +25,26 @@ module.exports = {
...
 
       try {
-        keysFinance = require('./keys-finance.json');
+        configFile = require('./keys-whitelist.json');
       } catch {
-        keysFinance = {
-          mainnet: [],
-          testnet: []
-        }
+        configFile = {
```
rg911, line 37: redundant here, 
the catch should capture & throw the exceptions here

src/views/modals/ModalTransactionCosignature/ModalTransactionCosignatureTs.ts
```diff
@@ -144,13 +145,31 @@ export class ModalTransactionCosignatureTs extends Vue {
...
+        if (this.hasMissSignatures && this.needsCosignature) {
+            return false;
+        }
+
+        // Check wether the Aggregate Bonded has inner Transfer with the current account address as a recipient.
+        const currentAddress = this.currentAccount.address;
+        const innerTransferTransaction = this.transaction.innerTransactions.find(
+            (innerTransaction) =>
+                innerTransaction.type === TransactionType.TRANSFER &&
+                (innerTransaction as TransferTransaction).recipientAddress?.plain() === currentAddress,
```
rg911, line 166: you can use the address.equal(...) perhaps? other wise, watch out the case

rg911 (changes requested):

vue.config.js
OlegMakarenko: > catch should capture
`console.error` added

> throw the exceptions here
Then it will be impossible to build the app without this json file

AdriaCarrera (approved):

rg911 (approved):

PR #1467, opened 2021-04-29
Use NetworkGenerationHash for network identification

### LINKED ISSUE #1412: Use NetworkGenerationHash for network identifier
Currently, the wallet uses NetworkType for identifying a network, I would be better to change to use genHash (as a unique hash) as a network identifier.

bassemmagdy:
fixes #1412


src/config/NetworkConfig.ts
```diff
@@ -149,10 +149,13 @@ export const defaultMainnetNetworkConfig: NetworkConfig = {
         { friendlyName: 'ngl-dual-605', roles: 2, url: 'http://ngl-dual-605.symbolblockchain.io:3000' },
     ],
 };
-
-const defaultNetworkConfig: Record<number, NetworkConfig> = {
-    152: defaultTestnetNetworkConfig,
-    104: defaultMainnetNetworkConfig,
+export const defaultGenerationHashes: Record<string, string> = {
```
fboucquez, line 152: Every time the defaultGenerationHashes is used, it's a code smell. The wallet is indirectly indexing by network type anyway.  network type -> default hash -> some data

src/core/database/storage/SettingsModelStorage.ts
```diff
@@ -43,7 +42,7 @@ export class SettingsModelStorage extends VersionedObjectStorage<Record<string,
                         profiles.map((name: string) => {
                             modified[name] = {
                                 ...modified[name],
-                                explorerUrl: networkConfig[NetworkType.TEST_NET].explorerUrl,
+                                explorerUrl: networkConfig[defaultGenerationHashes.TEST_NET].explorerUrl,
```
fboucquez, line 45: Hash needs to be a parameters, not a static field

src/core/utils/CommonHelpers.ts
```diff
@@ -47,4 +50,13 @@ export class CommonHelpers {
             await this.sleep(interval);
         }
     }
+
+    public static getGenerationHash(networkType: NetworkType): string {
```
fboucquez, line 54: If this is is used, your are indirectly indexing by network type

src/config/NetworkConfig.ts
```diff
@@ -149,10 +149,13 @@ export const defaultMainnetNetworkConfig: NetworkConfig = {
...
+    TEST_NET: '3B5E1FA6445653C971A50687E75E6D09FB30481055E3990C84B25E9222DC1155',
+    MAIN_NET: '57F7DA205008026C776CB6AED843393F04CD458E0AA2D9F1D5F31A402072B2D6',
+};
+const defaultNetworkConfig: Record<string, NetworkConfig> = {
+    '3B5E1FA6445653C971A50687E75E6D09FB30481055E3990C84B25E9222DC1155': defaultTestnetNetworkConfig,
```
fboucquez, line 157: If I add a second testnet like:
````
'AAA...': testnet2NetworkConfig,
````
would the wallet work?

fboucquez:
Hi @bassemmagdy , the main issue to solve is how the wallet can support multiples "Testnet" at the same time, like a testnet_1, testnet_2 (or a dev net). 

Everytime the generation hash is statically referenced, it's breaking that possibility. 
Also, each "network type" driven dropdown box (like profile creation, import, etc) also breaks this possibility. 

One example is this dropdown https://github.com/nemgrouplimited/symbol-desktop-wallet/blob/main/src/views/forms/FormProfileCreation/FormProfileCreation.vue#L50

The dropbox should be able to show multiple networks with the same network type.  Basically, those dropdown boxes should be driven by the list of known networks (like mainnet, testnet1, testnet2, and possible privateA, privateB, etc), not by network types. The list of known networks would be defined in the NetworkConfig.ts (`Objects.values(defaultNetworkConfig)`)

bassemmagdy:
updated in https://github.com/nemgrouplimited/symbol-desktop-wallet/pull/1467

PR #1470, opened 2021-04-30
Add support of multiple Opt-in transfers

OlegMakarenko:
- Display all opted-in NIS1 addresses related to the current account
- Display total Opt-in amount

image


segfaultxavi:
There should be no comma before "when" in the screenshot.
I haven't been able to find this string in the source so I cannot provide a patch, sorry.

rg911 (approved):

PR #1485, opened 2021-05-10
fix amount field bug while importing transaction from a qr-code

### LINKED ISSUE #1373: Redundant space in mosaics amount when imported from QR code with prepared invoice
**Steps to reproduce:**

1. Go to Invoice tab
2. Provide needed data into the fields on the page, specify mosaics amount with thousands (i.e. 1000 or 2000, 10000 etc.)
3. Download QR code
4. Import that QR code
5. Click Send button
**Expected result:** Mosaics amount is imported in the right format w/o redundant symbols
**Actual result:** There is a redundant space in mosaics amount preventing a user from successfully initiating a transaction

[Video](https://share.getcloudapp.com/llu5mkQ7)

bassemmagdy:
fixes #1373


surekabpm:
Tested, it is fixed and can be merged
image

rg911 (approved):

Issue #1433, opened 2021-04-15
Harvesting - node not visible on harvesting view after linking keys for multisig transaction


cryptoBeliever:
Choose node and link keys for multisig account and confirm on cosigners side.
Result:
Keys are linked but in Node Url input node is not presented
![image](https://user-images.githubusercontent.com/17684281/114910832-a3d84800-9e1e-11eb-85ca-e7a33634a835.png)

NikolaiB:
fixed

cryptoBeliever:
It doesn't work. I checked with 1 of 1 multi-sig.
Video: https://share.getcloudapp.com/geugpzKN

cryptoBeliever:
Fixed in integration.

PR #1502, opened 2021-05-24
fix connected nodes doesn't show up in node selector harvesting

### LINKED ISSUE #1106: The node connected to the wallet is not displayed in the delegation destination when setting the delegation harvest
in v0.14.0

in this case

bassemmagdy:
fixes #1106


rg911 (commented):
not sure how this PR fixes #1433, is it related to multisig?

bassemmagdy:
@rg911 it's not related to multisig as #1106 is the main issue and as a consequence for that #1433 got reported

AnthonyLaw:
Tested: #1433 

Screenshot 2021-06-09 at 10 09 18 PM

AdriaCarrera (approved):

PR #1506, opened 2021-05-27
A user should be able to open QRcode reader in offline mode 

### LINKED ISSUE #1472: Can't use the QR code reader in offline mode because there is a error message covering it
When using the offline wallet, there is no easy way to scan QR codes. If a person goes to their profile and tries to scan a QR code, they can't because an error message on the top of the app covers the "Import QR code" button. The error message is about the wallet not being able to connect to nodes.

bassemmagdy:
fix #1472

### DIFF
--- a/src/components/AccountLinks/AccountLinks.less
+++ b/src/components/AccountLinks/AccountLinks.less
@@ -28,6 +28,7 @@
     color: @purpleDark;
     display: flex;
     align-items: center;
+    z-index: 11;
 }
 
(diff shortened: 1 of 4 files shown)


AnthonyLaw:
look good, I'm able to click on `Import QR code` with alert pop-up.

Screenshot 2021-06-10 at 3 29 46 PM

AnthonyLaw (approved):

rg911 (approved):

PR #1516, opened 2021-06-03
fix aggregate bonded transaction error "Failure_Core_Future_Deadline"

### LINKED ISSUE #1490: Aggegate bonded - Failure_Core_Future_Deadline
Steps:

1. Change system clock +1h

image

Questions:

bassemmagdy:
- fix #1490 
- upgrading to newer alpha-sdk


src/views/modals/ModalTransactionConfirmation/ModalTransactionConfirmationTs.ts
```diff
@@ -466,9 +471,11 @@ export class ModalTransactionConfirmationTs extends Vue {
 
     private async ledgerAccountMultisigTransactionOnSigner(values) {
         const { ledgerService, currentPath, isOptinLedgerWallet, ledgerAccount, multisigAccount, stageTransactions, maxFee } = values;
+        await this.$store.dispatch('network/SET_TRANSACTION_DEADLINE', 42, { root: true });
```
rg911, line 474: why using 42 here? Max aggregate transaction deadline can be 48 hours

bassemmagdy: we've been discussing this, ended up decreasing them(LockHash and Aggregate) to 4h and 42h for a more safer option as the network might reject when the local time is slightly unsynchronized.

cryptoBeliever: @bassemmagdy
1) Why we used slightly unsynchronized delta for hashlock to 2h (6-4) and for Aggregate 6h (48-42). I don't think it has any sense.
2) Aren't we switched to node time instead local time? There should be no "slightly unsynchronized" then. So we should use 6 and 48 in my opinion. Do you see reason to not use it?

src/views/modals/ModalTransactionConfirmation/ModalTransactionConfirmationTs.ts
```diff
@@ -480,9 +487,11 @@ export class ModalTransactionConfirmationTs extends Vue {
             .then((signedAggregateTransaction) => {
                 return signedAggregateTransaction;
             });
+        await this.$store.dispatch('network/SET_TRANSACTION_DEADLINE', 4, { root: true });
```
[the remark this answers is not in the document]
cryptoBeliever: Same as in https://github.com/nemgrouplimited/symbol-desktop-wallet/pull/1516#discussion_r650365632

src/views/modals/ModalTransactionConfirmation/ModalTransactionConfirmationTs.ts
bassemmagdy: got your point and it makes sense updated them set deadlines to max again, thanks!

bassemmagdy:
updated here #1552 .

Issue #758, opened 2020-12-01
When put max value transaction details are not presented


cryptoBeliever:
Video: https://share.getcloudapp.com/Jruqw5v9

Steps:
1. Go send transaction view.
2. Put max allowed value into the amount. It's 9000000000000000.
3. Try to send a transaction. 

Result: Only input to put password is presented. There is no transaction details on the pop-up.
Expected result: Transaction details should be visible on pop-up.

When I'm sending 9000000000 then all is OK.

james-lin71:
This issue comes from symbol-sdk, when we pass large number as mosaic amount.
```
    /**
     * Compact higher and lower uint parts into a uint
     * @returns {number}
     */
    compact() {
        const result = format_1.RawUInt64.compact(this.toDTO());
        if (Array.isArray(result)) {
            throw new Error('Compacted value is greater than Number.Max_Value.');
        }
        return result;
    }
    ```
Once we pass a large number, this cause Number.Max_Value overflow issue by returning array. Maybe, we can consider the amount of mosaic value to be less than 9000000000, or any better solution for this one?

rg911:
Yep, the mosaic amount type on server is 64 byte unsigned integer, so compact will throw if it over the max integer. 
One option is to cast the long to numeric string, then call UInt64.fromNumericString() to pass to the sdk. Check the UInt64 class on sdk, it provides basic add / subtraction operations

coiki:
User Story: As a user sending mosaics, in the confirmation popup I want to see the details of my transaction regardless of any input, so I can confirm the details before entering my password and confirm the transaction.

Acceptance Criteria:

- AC1: When signing an outbound transaction, the details of the tx should always be shown.

cryptoBeliever:
Fixed in integration branch.

PR #1524, opened 2021-06-10
Add max amount validator.

AnthonyLaw:
### Fix
- added max validator on `amount` field, if the relative amount is more than `9,000,000,000` you are not able to send.

Resolved: #758 

Screenshot 2021-06-10 at 8 15 47 PM


src/core/validation/validators/MaxAmountValidator.ts
```diff
@@ -0,0 +1,29 @@
...
+@staticImplements<Validator>()
+export class MaxAmountValidator {
+    /**
+     * Validates the max amount of the mosaic supply. eg: 9,000,000,000
+     * @static
+     * @param {*} relativeAmount
+     * @returns {boolean}
+     */
+    public static validate(relativeAmount: string): boolean {
+        return Number(relativeAmount) <= 9_000_000_000;
```
OlegMakarenko, line 27: I think `9_000_000_000` should not be hardcoded.
Supply may differ depends on mosaic

AnthonyLaw: yeah, base on my understanding the validation rules only take care of the [standard mosaic validation](https://docs.symbolplatform.com/concepts/mosaic.html#initial-supply).

The rest will pass to the server for verification. @bassemmagdy correct me if I'm wrong.

But I will revisit the issue again.

AdriaCarrera (dismissed):

src/core/validation/validators/MaxAmountValidator.ts
bassemmagdy: @AnthonyLaw it shouldn't be hard coded IMO, furthermroe instead of limiting validation to `9000_000_000` which is not the max allowed isn't fixing the [issue](https://github.com/nemgrouplimited/symbol-desktop-wallet/issues/758) as the validation should be `<=9_000_000_000_000_000`, @cryptoBeliever what do you think?

cryptoBeliever: 8_999_999_999_000_000 is max value for mosaic creation.

AnthonyLaw: I will revisit this task.

AnthonyLaw: @bassemmagdy The reason used `9_000_000_000` for amount field is because relative amount.

cryptoBeliever: Hmm. Not sure if I understand. Max releative amount for coin with divisibility=0 is 8_999_999_999_000_000:

http://explorer.testnet.symboldev.network/mosaics/5D83F9AE97B99686

bassemmagdy (approved):

PR #1525, opened 2021-06-10
Fix namespace duration issue.

AnthonyLaw:
### Fix
- `deletedIn` will have 1-day extra compare to `expiredIn`, so when a namespace is expired, it should display `deletedIn` else it should show `expiredIn`

Resolved: #1525

### DIFF
--- a/src/services/NamespaceService.ts
+++ b/src/services/NamespaceService.ts
@@ -94,8 +94,8 @@ export class NamespaceService {
         const expiredIn = endHeight - namespaceGracePeriodBlocks - currentHeight;
         const deletedIn = endHeight - currentHeight;
         const expiration = expired
-            ? TimeHelpers.durationToRelativeTime(expiredIn, blockGenerationTargetTime)
-            : TimeHelpers.durationToRelativeTime(deletedIn, blockGenerationTargetTime);
+            ? TimeHelpers.durationToRelativeTime(deletedIn, blockGenerationTargetTime)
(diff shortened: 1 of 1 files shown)


src/services/NamespaceService.ts
```diff
@@ -94,8 +94,8 @@ export class NamespaceService {
         const expiredIn = endHeight - namespaceGracePeriodBlocks - currentHeight;
         const deletedIn = endHeight - currentHeight;
         const expiration = expired
-            ? TimeHelpers.durationToRelativeTime(expiredIn, blockGenerationTargetTime)
-            : TimeHelpers.durationToRelativeTime(deletedIn, blockGenerationTargetTime);
```
mm-s, line 98: Old logic: If expired then expiration=expired otherwise deleted;
New logic: If expired then expiration=deleted otherwise expired

It doesn't make sense except if you are inverting the logic deliberatelly?

AnthonyLaw: Hi @mm-s 
yes, I did it deliberately.

The server return `endHeight` is included namespace grace period blocks (2880 blocks around 1 day).

So the data on `deletedIn` has always 1 day (grace period) extra than `expiredIn `

Example:

Screenshot 2021-06-11 at 4 16 31 PM

when creating a new namespace it shows 30 days (it excluded grace period).

So when a namespace is not expired, it should show 30 days rather than 31 days

bassemmagdy (approved):

Issue #1473, opened 2021-05-01
Bug when entering amounts in offline mode


jabo38:
It seems like sending from an offline wallet doesn't work. When trying to enter a XYM amount instead of entering a whole number with decimal points a person has to enter absolute units.  Even then, it seems like the tx doesn't go through for some reason.

AnthonyLaw:
Hi Jeff.

>When sending XYM on the offline wallet from a non-multisig wallet

Are you mean to create ` Offline transactions` and announce it via CLI?

coiki:
@AnthonyLaw did we get any more info about this issue? or reproduction? Thanks.

AnthonyLaw:
Nope, waiting for more input.

jabo38:
when entering "10 XYM" in the offline tx mode and announcing it, the tx goes through as 0.000010 XYM.  Meaning if I really want to send 10 XEM, then I have to type in 10,000,000 which is quite scary if somebody had that much and made that mistake. 

I would duplicate this but I can't get any offline tx to work from a private key wallet. 

Not just this build, but every build since launch, I've not ever been able to make an offline tx work correctly from a single account.

AnthonyLaw:
Thanks for the input @jabo38 
I will take a look :)

coiki:
awesome!!

cryptoBeliever:
@AnthonyLaw it still doesn't work for me 

Steps to reproduce:

1. Turn off your internet connection, clear from local storage `networkCurrencyCache`, `networkCache`, `mosaicCache` in your browser.
2. Generate transaction QR

![image](https://user-images.githubusercontent.com/17684281/123487740-d4064900-d60e-11eb-8c0c-24ad37768343.png)

![image](https://user-images.githubusercontent.com/17684281/123487751-d8326680-d60e-11eb-887c-8b937a136109.png)

The first problem is that there is **Mosaic (1/1) Expired**. Expired should not be presented.

3. Turn on WiFi, login into wallet and try announce transaction. Example QR:

![signedTransaction5](https://user-images.githubusercontent.com/17684281/123488145-9c4bd100-d60f-11eb-8464-171881e89a2d.png)

Result: Transaction is not announced
Expected result: Transaction is announced

If I do entire process online generated QR allow me announce transaction.

AnthonyLaw:
Alright,  thanks for the test.
I will take a look on it

cryptoBeliever:
[...]

PR #1526, opened 2021-06-11
Fix offline tx.

AnthonyLaw:
### Fix
- Because of offline mode, getter store `mosaic/mosaics` return empty, can't get network currency's divisibility.

Note: 
1. To test this, **turn off** your internet connection, **clear** local storage on your browser.
2. created offline tx and download the QR code.
3. **Turn on** internet and log in to your wallet. 
4. selected `Import QR code` and use that QR code you downloaded.

Resolved: #1473


bassemmagdy (commented):
If we're adding `balanceMosaics` just because it contains network currency, I believe it'd be better to use `networkCurrency` from mosaic store rather than push the whole list of mosaics with balances.

AnthonyLaw:
> If we're adding balance mosaics just because it contains network currency, I believe it'd be better to use `networkCurrency` from mosaic store rather than push the whole list of mosaics with balances.

Alright, I will update it.
thanks 👍🏼

AnthonyLaw:
Close this. refer https://github.com/nemgrouplimited/symbol-desktop-wallet/pull/1538

PR #1529, opened 2021-06-14
Fix reconnect issue on home page.

### LINKED ISSUE #1523: Click on retry is not working on network connection alert.
Screenshot 2021-06-10 at 3 48 16 PM

Step:

1. Turn of wifi.
2. log in to the desktop wallet
3. node disconnect alert shows up.
4. turn back on the wifi and connected.
5. click on `Click to retry` nothing happen.

AnthonyLaw:
### Fix
- Reconnect is not working on the home page.

Resolved #1523 

Note:
@bassemmagdy I have removed the `if` statement to allow the home page able trigger `reconnect`, I not sure why disable trigger `reconnect` on home page, but tested all seem working fine.

### DIFF
--- a/src/views/layout/PageLayout/PageLayoutTs.ts
+++ b/src/views/layout/PageLayout/PageLayoutTs.ts
@@ -43,7 +43,7 @@ import Settings from '@/components/Settings/Settings.vue';
 import LogoutButton from '@/components/LogoutButton/LogoutButton.vue';
 import { URLInfo } from '@/core/utils/URLInfo';
 //@ts-ignore
-import ImportQRButton from '@/components/QRCode/ImportQRButton/ImportQRButton.vue';
+const ImportQRButton = () => import('@/components/QRCode/ImportQRButton/ImportQRButton.vue');
 import { AccountModel } from '@/core/database/entities/AccountModel';
 //@ts-ignore
 import AccountLinks from '@/components/AccountLinks/AccountLinks.vue';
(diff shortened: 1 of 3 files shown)


src/views/layout/PageLayout/PageLayoutTs.ts
```diff
@@ -275,8 +275,6 @@ export class PageLayoutTs extends Vue {
     }
 
     public reconnect() {
-        if (this.$route.fullPath !== '/home') {
-            this.$store.dispatch('network/CONNECT', { waitBetweenTrials: true });
-        }
+        this.$store.dispatch('network/CONNECT', { waitBetweenTrials: true });
```
bassemmagdy, line 278: I would suggest adding if condition in case the current path is`/login` so the wallet wouldn't try to connect to any of the nodes while users are not logged in yet.

AnthonyLaw: I checked the `reconnect` method only able triggers when `current_node_is_not_available` alert appears.

Does the login page will have this alert as well?

bassemmagdy: nope, not for the alert but we got a watcher in that class that could retry to re-connect the wallet even if users are not logged in.

AnthonyLaw:
Close this, refer https://github.com/nemgrouplimited/symbol-desktop-wallet/pull/1536

PR #1538, opened 2021-06-16
fix: divisibility problem for offline mode

AnthonyLaw:
### Fix
- Because of offline mode, getter store `mosaic/mosaics` return empty, can't get network currency's divisibility.

Note: 
1. To test this, **turn off** your internet connection, **clear** local storage on your browser.
2. created offline tx and download the QR code.
3. **Turn on** internet and log in to your wallet. 
4. selected `Import QR code` and use that QR code you downloaded.

Resolved: #1473

### DIFF
--- a/__mocks__/network.ts
+++ b/__mocks__/network.ts
@@ -1,0 +1,21 @@
+export const networkMock = {
+    fees: {
+        median: 10,
+        free: 0,
+        slow: 5,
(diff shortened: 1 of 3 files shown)


bassemmagdy (approved):

NikolaiB:
fixed

PR #1541, opened 2021-06-16
fix selected nodes not appearing in harvesting page

### LINKED ISSUE #1106: The node connected to the wallet is not displayed in the delegation destination when setting the delegation harvest

bassemmagdy:
fixes #1106


bassemmagdy:
rebased in #1548

PR #1546, opened 2021-06-18
Gh1090 improve transaction previews

bassemmagdy:


bassemmagdy:
A better rebased solution is here #1549 :)

PR #1548, opened 2021-06-18
fix harvesting node not showing while activated on different devices

### LINKED ISSUE #1106: The node connected to the wallet is not displayed in the delegation destination when setting the delegation harvest

bassemmagdy:
rebases #1541 
fixes #1106


AdriaCarrera (dismissed):

AnthonyLaw:
The fix is not working, after try on it.
I working some fix now.

AnthonyLaw (approved):

PR #1550, opened 2021-06-18
Fix password copy paste

### LINKED ISSUE #1167: Copy & Past of the password when creating the wallet should not be accepted.

bassemmagdy:
fixes #1167


src/components/PasswordInput/PasswordInput.vue
```diff
@@ -0,0 +1,18 @@
+<template>
+    <input
+        :value="value"
```
AnthonyLaw, line 3: I no sure are you still need the `value` as props.
because I didn't see you pass anything in `value` props

AnthonyLaw:
you may need to add a unit test for the component.

AnthonyLaw (approved):

PR #1552, opened 2021-06-21
update transactions deadline to use deadline service from sdk

### LINKED ISSUE #1490: Aggegate bonded - Failure_Core_Future_Deadline

bassemmagdy:
- fix #1490 
- update sdk version-> @1.0.2-alpha-202106161249


AdriaCarrera (approved):

AdriaCarrera (dismissed):

cryptoBeliever (commented):
Announcing any transaction doesn't work now.

cryptoBeliever:
After next try I got error:

image

When I announced aggregate bonded transaction. It happened once per a few tries.

package.json
```diff
@@ -132,7 +132,7 @@
         "symbol-hd-wallets": "0.14.1-alpha-202103051108",
         "symbol-paper-wallets": "^1.0.2",
         "symbol-qr-library": "0.14.1-alpha-202103081047",
-        "symbol-sdk": "1.0.2-alpha-202106231358",
+        "symbol-sdk": "1.0.2-alpha-202106161249",
```
AnthonyLaw, line 135: it's the issue in the alpha SDK  1.0.2-alpha-202106231358?

src/store/Network.ts
```diff
@@ -255,6 +275,7 @@ export default {
             commit('setConnected', false);
             commit('connectingToNodeInfo', undefined);
             commit('setFeesConfig', undefined);
+            commit('simpleTransactionDeadline', undefined);
```
AnthonyLaw, line 278: When reset happen, should reset 
`aggregateTransactionDeadline` and `hashLockTransactionDeadline` as well?

src/views/forms/FormTransactionBase/FormTransactionBase.ts
```diff
@@ -177,8 +182,13 @@ export class FormTransactionBase extends Vue {
     /**
      * it creates the deadlines for the transactions.
      */
-    protected createDeadline(): Deadline {
-        return Deadline.create(this.epochAdjustment);
+    protected createDeadline(deadlineInHours = 2): Deadline {
+        this.$store.dispatch('network/SET_TRANSACTION_DEADLINE', deadlineInHours);
```
AnthonyLaw, line 186: I believe you need `await` for this because it return promises?

src/views/forms/FormTransactionBase/FormTransactionBase.ts
```diff
@@ -177,8 +182,13 @@ export class FormTransactionBase extends Vue {
     /**
      * it creates the deadlines for the transactions.
      */
-    protected createDeadline(): Deadline {
-        return Deadline.create(this.epochAdjustment);
+    protected createDeadline(deadlineInHours = 2): Deadline {
```
AnthonyLaw, line 185: Just wondering, if can just apply the solution here. 
we do not need involved in the Store state.

Example

```ts
protected async createDeadline(deadlineInHours = 2): Deadline {
    return  await (await DeadlineService.create(repositoryFactory)).createDeadlineUsingServerTime(deadlineInHours);
}
```

```ts
// to call
await this.createDeadline(x hours)
```

src/views/forms/FormTransactionBase/FormTransactionBase.ts
bassemmagdy: It gets triggered on transaction form change for each field, so I don't think it needs to have the await function, so you'll already have the deadline.

src/store/Network.ts
```diff
@@ -538,6 +561,27 @@ export default {
             const peerNodes: NodeInfo[] = await nodeRepository.getNodePeers().toPromise();
             commit('peerNodes', _.uniqBy(peerNodes, 'host'));
         },
+        async SET_TRANSACTION_DEADLINE({ commit, getters }, deadlineInHours = 2) {
+            const repositoryFactory: RepositoryFactory = getters['repositoryFactory'] as RepositoryFactory;
+            const deadline = await (await DeadlineService.create(repositoryFactory)).createDeadlineUsingServerTime(deadlineInHours);
+            switch (deadlineInHours) {
+                case 2:
```
fboucquez, line 568: I would't put a dynamic value like a deadline in the store. The store is more for "infrastructure" objects, not to pass these kind of values. In general I don't like vuex, a lot harder to maintain and not type safe + adding dependencies to unrelated code

bassemmagdy: in order to maintain the different deadlines in case of I aggregates we had to keep them somewhere to make sure they're consistent.

bassemmagdy:
update in #1713

PR #1555, opened 2021-06-22
Fix harvesting not showing 25% of the beneficiary

AdriaCarrera:
Now receipts are reduced to add all the fees harvested per block, not only the first one


AnthonyLaw (approved):

AnthonyLaw:
Tested on my end.

Explorer -> http://explorer.testnet.symboldev.network/blocks/130681

Wallet
image

PR #1563, opened 2021-06-25
Use NetworkGenerationHash as main network identifier

### LINKED ISSUE #1412: Use NetworkGenerationHash for network identifier

bassemmagdy:
fixes #1412


bassemmagdy:
fixed in https://github.com/nemgrouplimited/symbol-desktop-wallet/pull/1568

PR #1565, opened 2021-06-28
fix: fee multiplier for offline tx

AnthonyLaw:
### Fix
 - added `0x` for network currency mosaic, if without `0x` it returns the wrong hex id.
 - use 100 as a min fee multiplier for offline tx, else it won't enough fees announce to the network.

read more: https://github.com/nemgrouplimited/symbol-desktop-wallet/issues/1473#issuecomment-868849985

Resolved: #1473

### DIFF
--- a/src/services/offline/MockModels.ts
+++ b/src/services/offline/MockModels.ts
@@ -33,7 +33,8 @@ export const OfflineGenerationHash = {
     [NetworkType.MAIN_NET]: networkConfig[NetworkType.MAIN_NET].networkConfigurationDefaults.generationHash,
 };
 
-export const OfflineTransactionFees = new TransactionFees(84587, 100, 1136363, 0, 0);
+// use 100 as min fee multiplier for offline transaction.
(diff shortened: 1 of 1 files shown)


src/services/offline/MockModels.ts
```diff
@@ -50,8 +51,8 @@ export const OfflineNetworkProperties = {
         new ChainProperties(
             true,
             true,
-            networkConfig[NetworkType.TEST_NET].networkConfigurationDefaults.currencyMosaicId,
-            networkConfig[NetworkType.TEST_NET].networkConfigurationDefaults.harvestingMosaicId,
+            '0x' + networkConfig[NetworkType.TEST_NET].networkConfigurationDefaults.currencyMosaicId,
```
rg911, line 54: why we need '0x' here?

AnthonyLaw: That is an issue for offline tx showing mosaic (xym) Expired.

because the mosaic id was `.substring(2)` in the `configurationStringAsString` method.

So added `0x` in the mock is to match mosaic id from chain network property
image

bassemmagdy (approved):

PR #1568, opened 2021-06-30
Private/User Network Support

fboucquez:
In order to support multi-network, I had to do a mid-size refactoring of the wallet.

1. The network configuration is now stored in the 'network' local storage. The network storage is not a cache anymore, but a place where the default and custom networks are stored. This includes any configuration required for offline operation. 
2. The networkConfig.ts static configuration is not referenced anymore, now the view needs to inject the current network/networkModel in order to know information about the network.  This model could be the stored default networks or a custom one added by the user.
3. There is a new 'Connect To New Network' button on the main screen. This goes to the new network form. The screen needs to be discussed and redesigned by the product and UI team.
4. The user provides name, node, explorer, and faucet url. Once the network is created the user can create a profile for the given network. The network configurations like properties and currencies are loaded and stored when querying the node. The info is stored in the 'network' storage.
5. The MockModels don't have any network-specific static/hardcoded data. Once the network is created, you can offline sign transactions for custom networks too. I don't like the offline repository factory and mocks. The offline feature shouldn't be calling the repository factory in the first place.
6. I have simplified the Network.ts store and NetworkService.ts.
7. Fixed issue with [AmountInput](https://github.com/nemgrouplimited/symbol-desktop-wallet/blob/main/src/components/AmountInput/AmountInputTs.ts#L63). The global network value was modified and the component wasn't watching the mosaicHex prop to update the validation.
7. Different improvements around types. 
8. Added bootstrap for private network testing. 

This the first implementation of the feature. We would need to discuss the look and feel of the new view.
This PR will require full regression testing too. 

Eventually, we should remove all the hardcoded XYM text from all the products (wallet, explorer, and faucet) and use whatever is the network currency name. If we want to go even further, we should remove the text "Symbol" or white label the products so they can be used in different private solutions. This is for a future PR.

### How to test:

To test a private network, you are going to create a local private test with bootstrap. The bootstrap custom preset has been already created in the desktop repo. 

1. Install bootstrap global "npm install -g symbol-bootstrap". Note, I could have installed bootstrap as a local dependency but we are using a global installation to not pollute the desktop dependencies. 
2. Checkout to this branch.
3  Run "npm run bootstrap-start" to start the local network
4. Run "npm run dev" to start the wallet server
5. Go to http://localhost:8080/ click on Connect a new network:
6. Fill out the form with:
- Name: My super private network
- Node URL: http://localhost:3000
- Explorer URL: http://localhost:90
- Faucet: URL http://localhost:100
7. Create a new profile for the given network.
8. Test the wallet :)! including offline

### TODOs:
1. Improve "Connect to new network" look and feel.
2. Move preLaunchOptin and nglFinanceBot to optionals fields in networkModel
3. Revisit logout reset state errors. Related to https://github.com/nemgrouplimited/symbol-desktop-wallet/issues/1575
4. Remove hardcoded XYM texts in code, labels, and translations. Use the network currency's 'ticker'

### For the future
1. Why offline tries to connect to the node. Can we avoid that? Is it possible to remove the offline repository and all the mocked models?
2. Simplify NodeService.getNodes, It seems there are redundant calls.  The node data is already loaded in NetworkService


bassemmagdy (approved):

cryptoBeliever:
Closing since we will not support private networks at this moment.
- https://github.com/symbol/desktop-wallet/pull/1568
- https://github.com/symbol/desktop-wallet/issues/1412

PR #1569, opened 2021-06-30
Fix CR

AnthonyLaw:
### Fix
- Display  "Encrypted message" if the message is encrypted.
- check on locale decimal separator before compare value.


cryptoBeliever:
- Encrypted message - working
- Validation is not working because it's always max amount as 8999999999 where I have mosaic with supply 8999999999000000000

image

cryptoBeliever:
After change we go back to original problem:

image

cryptoBeliever:
After fixes it seems fine :)

bassemmagdy (approved):

PR #1570, opened 2021-06-30
fix password issue

### LINKED ISSUE #1167: Copy & Past of the password when creating the wallet should not be accepted.

bassemmagdy:
fixes #1167


cryptoBeliever (approved):
From the testing point of view looks good.

Issue #1435, opened 2021-04-15
Harvesting - not able unlink VRF key link for multisig account


cryptoBeliever:
Steps:
1. Try unlink VRF key (trash icon) for multisig account:

![image](https://user-images.githubusercontent.com/17684281/114913127-1fd38f80-9e21-11eb-9c8b-95ed875cd01c.png)

2. Unlink VRF
![image](https://user-images.githubusercontent.com/17684281/114913199-35e15000-9e21-11eb-8cb1-5f2278c667b0.png)

Result: Only hash lock transaction is announced. No aggregate after lock confirmation.

cryptoBeliever:
It works fine now. Tested in integration branch.

PR #1583, opened 2021-07-07
Fix not able unlink VRF key link for multisig account 

AnthonyLaw:
### Fix
- It should use the signer account to sign the transaction rather than `tempTransactionSigner`

Resolved: #1435


cryptoBeliever:
Works fine in feature branch :+1:

bassemmagdy (approved):

Issue #1476, opened 2021-05-01
Problem activating delegated harvesting where all keys aren't linked. 


jabo38:
Steps to reproduce. 

1) I fund an account. 
2) Age in funds for a day that are over 10,000 but under 50m. 
3) I go to harvesting to link keys and initiate transaction 
4) Play around with wallet and start harvesting on another account 
5) Come back to the first account and see the following message

![image](https://user-images.githubusercontent.com/10395907/116793446-c8762600-aa94-11eb-92ca-ff68faea1663.png)

The first problem is that harvesting was activated on this device (using a Ledger just a few minutes ago). 

The second problem is that seemingly not all the keys are linked? 

![image](https://user-images.githubusercontent.com/10395907/116793469-ea6fa880-aa94-11eb-828f-a066e12a593e.png)

Not sure if that is part of the problem or not, but after manually entering in the node I used last time into the node box, I'm then given the only option to unlink my keys and then to later relink them. 

![image](https://user-images.githubusercontent.com/10395907/116793507-1db23780-aa95-11eb-9607-6927d2100a6f.png)

Unlinking and relinking works.

bassemmagdy:
@jabo38 would you provide proper steps to reproduce?

jabo38:
I did pretty much write the steps to reproduce.

bassemmagdy:
@jabo38 Did you activate your harvesting on a different account or is it the same account in a different profile?

jabo38:
All the screen shots are for the same account in the same profile on the same computer and same desktop app.

cryptoBeliever:
I can reproduce problem with saving keys in this way:
- go to harvesting page (for account that have > 10000 XYM and importance > 0)
- click "Link keys" (transaction will be sent)
- do some interruption, like: change node, logout. After transaction will be confirmed you will end up with not saved private keys.

cryptoBeliever:
@AnthonyLaw not sure if it should be seperate issue or belongs to this one. But it's very similar (this time wrong node is selected instead missing keys). If it should be reported as seperated one please let me know.

Steps:
1. Create empty profile
2. Import account with balance for which we will start harvesting.
3.  Go to harvesting page (by default will be selected current node - X)
4. Switch harvesting node to Y
5. Start harvesting, before transaction will be confirmed switch to other profile and switch current node to Z
6. Wait for harvesting transaction confirmation
7. After confirmation done switch to harvesting view and selected node will be current node (Z) instead correct node (Y)

[...]

PR #1594, opened 2021-07-15
Prevent private key lose during Account key link

AnthonyLaw:
### Fix
- To store Remote and VRF key info before announcing the transaction, to prevent listener or web socket disconnect and lose the private key.
- Pre-store `selectedHarvestingNodes` when announcing link txs, and removed if supplement node pubkey is not exist

Resolved: #1476


AnthonyLaw:
To continue to fix this issue, https://github.com/symbol/desktop-wallet/issues/1476#issuecomment-888589927
we need to fix the SDK first. https://github.com/symbol/symbol-sdk-typescript-javascript/issues/806

cryptoBeliever:
@AnthonyLaw looks like harvesting for multisig stopped working:

https://share.getcloudapp.com/12uveLeR

src/store/Harvesting.ts
```diff
@@ -328,40 +353,127 @@ export default {
             { commit },
             { accountAddress, selectedHarvestingNode }: { accountAddress: string; selectedHarvestingNode: NodeModel },
         ) {
-            const harvestingService = new HarvestingService();
+            console.log('selectedHarvestingNode :>> ', selectedHarvestingNode);
```
bassemmagdy, line 356: could you remove logs?

src/store/Harvesting.ts
```diff
@@ -328,40 +353,127 @@ export default {
...
+                newEncVrfPrivateKey,
+                newRemotePublicKey,
+                newVrfPublicKey,
+                encRemotePrivateKey,
+                encVrfPrivateKey,
+            } = currentSignerHarvestingModel;
+
+            // If remote key link not exist from network or local
+            // Set local encRemotePrivateKey to null
+            if (!linked || !newRemotePublicKey) {
```
bassemmagdy, line 443: should be `&&` instead of `||`, no?

AnthonyLaw: I believe use `||` is more suitable in this situation.
because if 1 of them is empty, just set to `null`

bassemmagdy (approved):

PR #1631, opened 2021-08-04
added address next to alias in confirmation screen for clarity 

### LINKED ISSUE #1151: Could add address next to alias in confirmation screen for clarity.
**Is your feature request related to a problem? Please describe.**
No problem really, just a very small thing that could make the wallet nicer

**Describe the solution you'd like**
When I send a tx by alias, it could be nice to confirm the address in the confirmation screen, just in case I mistyped or something. 

**Describe alternatives you've considered**
I made a mockup below of how it could work. 

image

bassemmagdy:
fixes #1151

### DIFF
--- a/src/store/Namespace.ts
+++ b/src/store/Namespace.ts
@@ -149,6 +149,7 @@ export default {
             const linkedAddress = await getLinkedAccountPromise;
 
             commit('linkedAddress', linkedAddress);
+            return linkedAddress;
         },
 
         SIGNER_CHANGED({ commit, rootGetters, getters }) {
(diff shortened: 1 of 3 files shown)


src/store/Namespace.ts
```diff
@@ -149,6 +149,7 @@ export default {
             const linkedAddress = await getLinkedAccountPromise;
 
             commit('linkedAddress', linkedAddress);
+            return linkedAddress;
```
AnthonyLaw, line 152: I don't think you need a return here, because it already save in the `linkedAddress` state

AnthonyLaw (approved):

PR #1648, opened 2021-08-18
show the correct signer for aggregate bonded transactions in dashboard

### LINKED ISSUE #1564: Dashboard - For aggregate tx "From" is presented as initializing cosigner instead multisig account
Steps to reproduce:

1. Create multi-sig account (in this example 2 of 2 multi-sig) and not related wallet W1.

M TDJZJWJCW7W7EZSOHPWL4BBCYBVUJEP7WN52KCA
	C1 TB2Z6E5VCEFQQN4DTZNYDPJYRFDSYCSQYYLPVXY
	C2 TASS5DO72MEOWHZZMTTIOPIVA5F7W4QZGYKRUNA

W1 TBYZ5FOQCJB4GRDUOFJWWPMQYQWHGRBLCVWV2KI

2. Send multisig (aggregate) transaction from multisig M using cosigner C1 to wallet W1.
3. Sign transaction using cosigner C2.

Result:

On dashboard you can see From as TB2Z6E5VCEFQQN4DTZNYDPJYRFDSYCSQYYLPVXY which is cosigner instead multisig account:

image

In details "From" is presented correctly:

image

Expected result:
Present on dashboard correct "From" value.

If this is a limitation of REST API we should think of changing it.

bassemmagdy:
fix #1564


src/components/TransactionList/TransactionRow/TransactionRowTs.ts
```diff
@@ -412,11 +412,12 @@ export class TransactionRowTs extends Vue {
     @Watch('transaction', { immediate: true })
     private async fetchTransaction() {
         if (
-            this.transaction instanceof AggregateTransaction &&
-            this.transaction.type === TransactionType.AGGREGATE_BONDED &&
-            this.hasMissSignatures
+            (this.transaction instanceof AggregateTransaction &&
+                this.transaction.type === TransactionType.AGGREGATE_BONDED &&
+                this.hasMissSignatures) ||
+            this.transaction instanceof AggregateTransaction
```
yilmazbahadir, line 418: It seems like this is equivalent to the following, isn't it?
```
if (this.transaction instanceof AggregateTransaction) 
```

src/components/TransactionList/TransactionRow/TransactionRowTs.ts
```diff
@@ -124,7 +124,7 @@ export class TransactionRowTs extends Vue {
      * Current transaction Details
      * @type {AggregateTransaction}
      */
-    private aggregateTransactionDetails: AggregateTransaction;
+    private aggregateTransactionDetails: AggregateTransaction = null;
```
yilmazbahadir, line 127: Can you elaborate on this change?

src/components/AddressDisplay/AddressDisplayTs.ts
```diff
@@ -45,13 +45,14 @@ export class AddressDisplayTs extends Vue {
...
-    protected async loadDetails(): Promise<void> {
+    @Watch('address', { immediate: true })
+    async loadDetails(): Promise<void> {
         this.descriptor = '';
         if (!this.address) {
             return;
         }
         // in case of normal transfer, display pretty address
-        if (this.address instanceof Address) {
+        if (this.address && this.address instanceof Address) {
```
yilmazbahadir, line 55: Do we really need this change here? (pls check line 51, already checking if the address is truthy)

src/components/ActionDisplay/ActionDisplayTs.ts
```diff
@@ -72,4 +81,10 @@ export class ActionDisplayTs extends Vue {
 
         return i18n.t(`transaction_descriptor_${this.transaction.type}${this.isOptinPayoutTransaction ? '_optin' : ''}`);
     }
+    @Watch('aggregateTransactionSenderAddress', { immediate: true })
```
yilmazbahadir, line 84: Should we put a new line before this line?

src/components/ActionDisplay/ActionDisplayTs.ts
```diff
@@ -62,6 +70,7 @@ export class ActionDisplayTs extends Vue {
      */
     protected needsCosignature: boolean = false;
 
+    protected hasAggregateBondedSigner: boolean = false;
     /**
```
yilmazbahadir, line 74: Should we put a new line before this line?

src/main.ts
```diff
@@ -66,6 +67,7 @@ Vue.directive('click-outside', clickOutsideDirective);
 /// end-region directives
 
 const app = new Vue({
+    // @ts-ignore
```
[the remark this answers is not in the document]
bassemmagdy: vue-router causes some conflicts with typescript version which produce console errors sometimes, here is a reference https://github.com/vuejs/vue-router/issues/3019

src/components/TransactionList/TransactionRow/TransactionRowTs.ts
bassemmagdy: the second case ignores checking for missing signature.

src/components/TransactionList/TransactionRow/TransactionRowTs.ts
bassemmagdy: I needed to reference that in `TransactionRow.vue` file so It requires to have a value declared.

src/components/TransactionList/TransactionRow/TransactionRowTs.ts
yilmazbahadir: This `if` check will result in `true` if `this.transaction instanceof AggregateTransaction` is `true` (because of the second check - right-hand side of the 'or' operator), right? If so what's the point of having the first part of the `||` here? In any case when the transaction is instance of AggregateTransaction then it will return true then this is equivalent to the following if that makes sense?
```
if (this.transaction instanceof AggregateTransaction) 
```

src/main.ts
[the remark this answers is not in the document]
yilmazbahadir: Interesting that it happens sometimes, perhaps that's the reason, I couldn't see any errors in my local console, could you provide any logs regarding this issue if possible? Keen to understand the issue better, that'd be nice if we can find a better solution rather than suppressing it.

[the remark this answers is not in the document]
bassemmagdy: Screen Shot 2021-08-24 at 2 28 06 PM

yilmazbahadir (approved):

PR #1653, opened 2021-08-25
fix: using AGGREGATE_COMPLETE on aggregate txs when requiredCosignatures is 1

yilmazbahadir:
Especially in FormMultisigAccountModificationTransaction and FormPersistentDelegationRequestTransaction forms some cases were uncovered; refactored to use `requiredCosignatures` in the base class and refactored harvesting form to create agg complete tx when possible.


src/views/forms/FormPersistentDelegationRequestTransaction/FormPersistentDelegationRequestTransactionTs.ts
```diff
@@ -624,6 +609,37 @@ export class FormPersistentDelegationRequestTransactionTs extends FormTransactio
         );
     }
 
+    private signAndAnnounceTransactions(
```
AnthonyLaw, line 612: it will be good to add a description for this method.

yilmazbahadir: Added.

AnthonyLaw (approved):

Issue #1677, opened 2021-08-30
Aggregate complete - invalid date of send transaction


cryptoBeliever:
Steps:
1. Init aggregate complete.
2. Check transaction details

Result:
In transaction history list date is presented as -46h comparing real date.

Explanation:
Aggregate complete is sent with deadline 2h but when wallet calculate date there is subtracted 48 hours (as for bonded).

![image](https://user-images.githubusercontent.com/17684281/131378554-5e7f142f-62e5-4ab9-988c-419bc0467b67.png)

This transaction was send: **2021-08-30 19:11:09**

Issue #1675, opened 2021-08-30
Multisig account transfer view - mosaics not visible


cryptoBeliever:
Steps:

1. Create multisig.
2. Create mosaics for multisig and for cosigners.
3. Go to transfer view
4. Switch account to multisig

Result: Mosaics available for multisig are not visible

Video: https://www.loom.com/share/feb80de98cb24098b479977d359663f6

cryptoBeliever:
Fixed.

Issue #1577, opened 2021-07-05
When preparing a tx and a previous tx is confirmed, the send fields on a new tx are reset.


jabo38:
When I announce a tx, as soon as it is sent, the fields I fill out like "to" or "amount" or "message" get reset to default when the tx is announced. 

I go to fill in more information, and if the previous tx is is confirmed, then the fields are reset to default. 

This happens a lot with multisig txs when trying to initiate one after another.

cryptoBeliever:
Looks fine in dev.

cryptoBeliever:
@bassemmagdy changes from https://github.com/symbol/desktop-wallet/pull/1680 regressed this one.

cryptoBeliever:
Will be fixed in https://github.com/symbol/desktop-wallet/issues/1683

PR #1680, opened 2021-08-30
regression bug fixing

bassemmagdy:
Fixing:
- #1675
- #1677

### DIFF
--- a/src/views/forms/FormTransferTransaction/FormTransferTransactionTs.ts
+++ b/src/views/forms/FormTransferTransaction/FormTransferTransactionTs.ts
@@ -602,6 +602,7 @@ export class FormTransferTransactionTs extends FormTransactionBase {
     @Watch('selectedSigner')
     onSelectedSignerChange() {
         this.formItems.signerAddress = this.selectedSigner.address.plain();
+        this.resetForm();
     }
 
     /**
--- a/src/components/TransactionList/TransactionRow/TransactionRowTs.ts
+++ b/src/components/TransactionList/TransactionRow/TransactionRowTs.ts
@@ -466,7 +466,7 @@ export class TransactionRowTs extends Vue {
     }
 
     public get date(): string {
-        if (this.transaction instanceof AggregateTransaction) {
+        if (this.transaction.type === TransactionType.AGGREGATE_BONDED) {
             return TimeHelpers.getTransactionDate(this.transaction.deadline, 48, this.networkConfiguration.epochAdjustment);
         } else if (this.transaction.type === TransactionType.HASH_LOCK) {
             return TimeHelpers.getTransactionDate(this.transaction.deadline, 6, this.networkConfiguration.epochAdjustment);


AnthonyLaw (approved):

src/views/forms/FormTransferTransaction/FormTransferTransactionTs.ts
```diff
@@ -602,6 +602,7 @@ export class FormTransferTransactionTs extends FormTransactionBase {
     @Watch('selectedSigner')
     onSelectedSignerChange() {
         this.formItems.signerAddress = this.selectedSigner.address.plain();
+        this.resetForm();
```
yilmazbahadir, line 605: Wouldn't this break this change? => https://github.com/symbol/desktop-wallet/pull/1642
Should we just update the necessary parts instead?

bassemmagdy: nope, it wouldn't break it. After having this sticky account feature it works as expected.

cryptoBeliever: After sending from multisig "From" selector stays selected on previously choosen multisig.

cryptoBeliever (approved):
Works fine :+1:

src/views/forms/FormTransferTransaction/FormTransferTransactionTs.ts
yilmazbahadir: Let's say as a user, I filled the transfer form and then noticed that I've chosen the wrong account, currently, all the other fields remain the same when I changed the signer but with this change, I'll have to re-enter all the information since it'll reset the form, right? I'm not sure if that's the behaviour we want to have in the transfer form tbh, @cryptoBeliever @coiki any ideas?

bassemmagdy: As we need the form to be reactive especially when it comes to the held mosaics in each account, I believe we need to reset fields on the change of signer.

cryptoBeliever: @yilmazbahadir you are right. I thought about other case here. 
@bassemmagdy would be possible to refresh only mosaics?

yilmazbahadir: we can only reset the held mosaics (dropdown and maybe the amount - unsure) rather than resetting the form, right? (If it won't break the reactivity ofc)

https://github.com/symbol/desktop-wallet/pull/1680#discussion_r699210817
> Should we just update the necessary parts instead?

cryptoBeliever: Form is cleaned after tx confirm. So we have regression with this fix -> https://github.com/symbol/desktop-wallet/issues/1577

Issue #1679, opened 2021-08-30
Mosaics - not able to edit supply or link alias for mosaic owned by multisig


cryptoBeliever:
Steps:
1. Create multisig
2. Create mosaic with mutable supply for multisig account
3. Go to mosaics list

Result: There is no option to edit supply or link alias for owned mosaic
Expected result: There is option to edit supply and link alias. Same as for regular accounts.

Video: https://www.loom.com/share/c001508075284ed69f8e7a12da63792d

cryptoBeliever:
Fixed.

PR #1681, opened 2021-08-30
fix: multisig account owned mosaic actions visibility issue fixed

yilmazbahadir:
fixes https://github.com/symbol/desktop-wallet/issues/1679

Was filtering out only the currentAccount(changed to currentSigner now) for the actions on the mosaic list therefore missing the multisig accounts, looks like the issue was always there but became more visible with the 'sticky' signer selection fix.

### DIFF
--- a/src/components/TableDisplay/TableDisplayTs.ts
+++ b/src/components/TableDisplay/TableDisplayTs.ts
@@ -51,7 +51,6 @@ import ModalMetadataDisplay from '@/views/modals/ModalMetadataDisplay/ModalMetad
 import { NamespaceModel } from '@/core/database/entities/NamespaceModel';
 import { MosaicModel } from '@/core/database/entities/MosaicModel';
 import { NetworkConfigurationModel } from '@/core/database/entities/NetworkConfigurationModel';
-import { AccountModel } from '@/core/database/entities/AccountModel';
 import { Signer } from '@/store/Account';
 // @ts-ignore
 import SignerFilter from '@/components/SignerFilter/SignerFilter.vue';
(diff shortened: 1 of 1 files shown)


src/components/TableDisplay/TableDisplayTs.ts
```diff
@@ -14,9 +14,19 @@
  *
  */
 // external dependencies
-import { Component, Prop, Vue, Watch } from 'vue-property-decorator';
-import { mapGetters } from 'vuex';
-import { Address, AliasAction, MosaicId, NamespaceId } from 'symbol-sdk';
+// @ts-ignore
+import ButtonAdd from '@/components/ButtonAdd/ButtonAdd';
```
AnthonyLaw, line 18: Do we need the `.vue` extension?

src/components/TableDisplay/TableDisplayTs.ts
```diff
@@ -14,9 +14,19 @@
...
+// @ts-ignore
+import ButtonRefresh from '@/components/ButtonRefresh/ButtonRefresh';
```
AnthonyLaw, line 20: same as above.

src/components/TableDisplay/TableDisplayTs.ts
AnthonyLaw, line 18:
```suggestion
import ButtonAdd from '@/components/ButtonAdd/ButtonAdd.vue';
```

src/components/TableDisplay/TableDisplayTs.ts
AnthonyLaw, line 20:
```suggestion
import ButtonRefresh from '@/components/ButtonRefresh/ButtonRefresh.vue';
```

yilmazbahadir:
@AnthonyLaw Just used the `organize imports` feature to remove unused and reorganize, so there is not an actual change in the import parts.

AnthonyLaw (approved):

cryptoBeliever (approved):
Tested and works fine :+1:

PR #1682, opened 2021-08-31
separating mosaics list re-setting

### LINKED ISSUE #1675: Multisig account transfer view - mosaics not visible

bassemmagdy:
- fixes #1675

### DIFF
--- a/src/views/forms/FormTransferTransaction/FormTransferTransactionTs.ts
+++ b/src/views/forms/FormTransferTransaction/FormTransferTransactionTs.ts
@@ -281,13 +281,17 @@ export class FormTransferTransactionTs extends FormTransactionBase {
 
         // transaction details passed via router
         this.importTransaction = this.$route.params.transaction || this.importedTransaction ? true : false;
-        this.resetMosaicsAndTriggerChange(this.importTransaction);
+        this.resetMosaics(this.importTransaction);
+        this.triggerChange();
     }
 
     /**
      * Resetting mosaics list when signer is changed
+     * @param {boolean} importedTransaction checks if transaction is imported
+     * @param {string} triggerChange checks if triggerChange() needs to be called
+     * @return {void}
      */
-    private resetMosaicsAndTriggerChange(importedTransaction: boolean): void {
+    private resetMosaics(importedTransaction: boolean, triggerChange = false): void {
         this.mosaicInputsManager = MosaicInputsManager.initialize(this.currentMosaicList());
         // - reset attached mosaics
         this.formItems.attachedMosaics = [];
@@ -315,7 +319,9 @@ export class FormTransferTransactionTs extends FormTransactionBase {
                 Vue.set(this.formItems.attachedMosaics, index, attachedMosaic);
             });
         }
-        this.triggerChange();
+        if (triggerChange) {
+            this.triggerChange();
+        }
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/views/forms/FormTransferTransaction/FormTransferTransactionTs.ts
```diff
@@ -278,6 +273,27 @@ export class FormTransferTransactionTs extends FormTransactionBase {
...
+        // transaction details passed via router
+        this.importTransaction = this.$route.params.transaction || this.importedTransaction ? true : false;
+        this.resetMosaics(this.importTransaction);
+        this.triggerChange();
+    }
+
+    /**
+     * Resetting mosaics list when signer is changed
+     */
+    private resetMosaics(importedTransaction: boolean): void {
```
yilmazbahadir, line 291: Should `resetMosaics` call `triggerChanges` conditionally? In the case of a selected mosaic change after the reset, then the fee calculation might change, not 100% sure though.

If so maybe having a second parameter `triggerChanges = false(default)` should be added; and at the end of the method, we can have the following.

```
if (triggerChanges) {
    this.triggerChange();
}
```

bassemmagdy: In that case, Do we need the condition and an extra param? we could move the` triggerChange()` call inside `resetMosaics()` and remove it from `resetForm` instead.

yilmazbahadir: Yes, that also works, in that case I'd rename the method to `resetMosaicsAndTriggerChange`.

yilmazbahadir (approved):
@cryptoBeliever could you please approve if testing looks OK?

cryptoBeliever:
@yilmazbahadir @bassemmagdy 

Would be possible not reset mosaic selection (but change select) and amount if account to which we changing also has such mosaic?

I mean we have two accounts:
- Account A with mosaics X, Y, Z
- Account B with mosaics X, Z

Before changing from account A to B we had selected mosaic Z. So after changing to B we should have on select only X and Z but we should have still selected Z and amount will stay.

Resetting amount is especially problematic when transaction got confirmed (video 1:08).

Video: https://share.getcloudapp.com/lluooAl8

bassemmagdy:
@cryptoBeliever @yilmazbahadir is it worth adding an extra dependency into the store to save previously selected one from first signer?

cryptoBeliever:
@bassemmagdy 
Just wonder what is the reason behind resetting the form when notification about transaction confirmation arrives?
This is most strange for the user because he already filling the form again to send a new transaction and mosaics got cleared.
Maybe we can reset mosaics component only on changing sender and not receiving tx confirmation?

bassemmagdy:
@cryptoBeliever added more restrictions to the [watcher](https://github.com/symbol/desktop-wallet/blob/811562f2a9b2819752248f0e77679c3f43bc793a/src/views/forms/FormTransferTransaction/FormTransferTransactionTs.ts#L615) as it was getting executed without the signer getting changed, please re-test the updates.

cryptoBeliever:
@bassemmagdy now it's problem that not all mosaics are presented for user:
https://share.getcloudapp.com/bLuqdDpj

Also problem with resetting on confirmation is global. Here example for namespace creation (on branch from PR):
https://share.getcloudapp.com/YEuOPmv0

I think we can create a separate issue for that and revert the latest changes if it's problematic to fix that.

cryptoBeliever:
After discussion I created two separate issues for that:
- https://github.com/symbol/desktop-wallet/issues/1683
- https://github.com/symbol/desktop-wallet/issues/1684

That's why I'm closing this PR but part of code can be reused.

PR #1685, opened 2021-09-01
fix: transfer and metadata tx form multi-level multi-sig agg complete fix

yilmazbahadir:
Calculates correct required cosignatures needed and creates `aggregate complete` tx if there is no multi-sig signer in the tree(_between the levels from selected signer to the current account_) that has `minApproval` greater than 1.

Known issue with the current implementation => https://github.com/symbol/desktop-wallet/issues/1687
Will use a tree-like data structure to implement it.


cryptoBeliever (approved):
Works fine :+1:

bassemmagdy (approved):

PR #1689, opened 2021-09-02
fix: multisig account graph info stored value fix

yilmazbahadir:
`multisigAccountGraphInfo` store value was set to an array of arrays at one line and flat array at another, I think this PR will fix it to use the same value.

image

### DIFF
--- a/src/store/Account.ts
+++ b/src/store/Account.ts
@@ -527,8 +527,9 @@ export default {
                     map((g) => {
                         // sorted array to be represented in multisig tree
                         commit('multisigAccountGraph', g.multisigEntries);
-                        commit('multisigAccountGraphInfo', MultisigService.getMultisigGraphArraySorted(g.multisigEntries));
-                        return MultisigService.getMultisigInfoFromMultisigGraphInfo(g);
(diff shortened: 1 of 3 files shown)


cryptoBeliever (dismissed):
Problem fixed :+1:

bassemmagdy:
for unifying the 2D Array type, I guess we need to update it in a couple of other places:
- https://github.com/symbol/desktop-wallet/blob/4860bc61e0cc037906a4934fd2d62d797b57fb24/src/store/Account.ts#L421
- https://github.com/symbol/desktop-wallet/blob/4860bc61e0cc037906a4934fd2d62d797b57fb24/src/store/Account.ts#L535

yilmazbahadir:
> for unifying the 2D Array type, I guess we need to update it in a couple of other places:
> 
> * https://github.com/symbol/desktop-wallet/blob/4860bc61e0cc037906a4934fd2d62d797b57fb24/src/store/Account.ts#L421
> * https://github.com/symbol/desktop-wallet/blob/4860bc61e0cc037906a4934fd2d62d797b57fb24/src/store/Account.ts#L535

This change is 2D -> 1D already, why would we need it here?

bassemmagdy (approved):

PR #1693, opened 2021-09-03
Ledger harvesting

AnthonyLaw:
### Fix
- Store vrf and remote key info before announce.
- unable to unlink a single key, caused by some [reactive properties](https://vuejs.org/v2/guide/reactivity.html#Declaring-Reactive-Properties) undefine, and assign a default value.

### DIFF
--- a/src/views/forms/FormPersistentDelegationRequestTransaction/FormPersistentDelegationRequestTransactionTs.ts
+++ b/src/views/forms/FormPersistentDelegationRequestTransaction/FormPersistentDelegationRequestTransactionTs.ts
@@ -170,8 +170,8 @@ export class FormPersistentDelegationRequestTransactionTs extends FormTransactio
 
     private tempTransactionSigner: TransactionSigner;
     private tempAccount: Account;
-    public vrfPrivateKeyTemp: string;
-    public remotePrivateKeyTemp: string;
+    public vrfPrivateKeyTemp: string = '';
+    public remotePrivateKeyTemp: string = '';
 
     /**
      * Panel tab management getters/setters
@@ -238,7 +238,7 @@ export class FormPersistentDelegationRequestTransactionTs extends FormTransactio
      * @var {string}
      */
     private type: string;
-    private password: string;
+    private password: string = '';
     private linkIcon: string = officialIcons.publicChain;
     private linking = false;
     private showModalImportKey: boolean = false;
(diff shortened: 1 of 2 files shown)


src/views/modals/ModalTransactionConfirmation/ModalTransactionConfirmationTs.ts
```diff
@@ -742,6 +760,29 @@ export class ModalTransactionConfirmationTs extends Vue {
...
+            this.saveRemoteKeyInfo(
+                accountAddress,
+                Crypto.encrypt(this.command.newRemoteAccount.privateKey, this.command.password),
+                this.command.newRemoteAccount.publicKey,
+            );
+        }
+
+        // // pre-store selected harvesting node in local
+        this.saveHarvestingNode(accountAddress, this.command.formItems.nodeModel);
+
```
yilmazbahadir, line 785: This block looks like the block (L579-601), could we extract to a method and reuse that method?

src/views/forms/FormPersistentDelegationRequestTransaction/FormPersistentDelegationRequestTransactionTs.ts
```diff
@@ -170,8 +170,8 @@ export class FormPersistentDelegationRequestTransactionTs extends FormTransactio
 
     private tempTransactionSigner: TransactionSigner;
     private tempAccount: Account;
-    public vrfPrivateKeyTemp: string;
-    public remotePrivateKeyTemp: string;
+    public vrfPrivateKeyTemp: string = '';
```
yilmazbahadir, line 173: Why are we initializing the temp keys?

yilmazbahadir: I guess that's the reason
> unable to unlink a single key, caused by some reactive properties undefine, and assign a default value.

src/views/modals/ModalTransactionConfirmation/ModalTransactionConfirmationTs.ts
AnthonyLaw: yes, that is possible. Was thinking about where should we put this method.
HarvesrtingServices?

yilmazbahadir: I think a private method within the same file would be enough for now.

cryptoBeliever:
It works now :+1: 
Checked:
- Ledger single account starting and stopping harvesting + manually removing links from "Key links" tab
- Ledger multi-sig account starting and stopping harvesting + manually removing links from "Key links" tab
- Regualr profile single harvesting
- Regular profile multisig harvesting

yilmazbahadir (approved):

PR #1699, opened 2021-09-09
Aggegate bonded - Failure_Core_Future_Deadline

### LINKED ISSUE #1490: Aggegate bonded - Failure_Core_Future_Deadline

bassemmagdy:
fixes #1490

### DIFF
--- a/__tests__/views/forms/FormTransferTransaction.spec.ts
+++ b/__tests__/views/forms/FormTransferTransaction.spec.ts
@@ -125,8 +125,7 @@ describe('FormTransferTransaction', () => {
         });
 
         // act
-        const tx = vm.getTransactions();
-
+        const tx = await vm.getTransactions();
         // assert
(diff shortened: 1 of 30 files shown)


cryptoBeliever:
@bassemmagdy looks like the number of API calls to the /time endpoint is too big. Please check video:
https://share.getcloudapp.com/BluxeB7B

Maybe wallet should have only time synchronization periodically called like once per minute?

yilmazbahadir:
Can we just call the time api on the connected node manually, get the server time and hold the time difference in the state in milliseconds and create the deadline by using that difference when constructing the transaction?

bassemmagdy:
do you mean we create we keep the deadlines objects in store till we need them in transaction building?

bassemmagdy:
update in #1713

PR #1713, opened 2021-09-23
Gh1490 deadlines issue

### LINKED ISSUE #1490: Aggegate bonded - Failure_Core_Future_Deadline

bassemmagdy:
- fixes #1490 by finding the time difference between local machine and server and then use it to create the deadline object

### DIFF
--- a/src/services/TransactionCommand.ts
+++ b/src/services/TransactionCommand.ts
@@ -34,6 +34,7 @@ import { Observable, of } from 'rxjs';
 import { AccountTransactionSigner, TransactionAnnouncerService, TransactionSigner } from '@/services/TransactionAnnouncerService';
 import { BroadcastResult } from '@/core/transactions/BroadcastResult';
 import { flatMap, map } from 'rxjs/operators';
+import { AppStore } from '@/app/AppStore';
 
 export enum TransactionCommandMode {
     SIMPLE = 'SIMPLE',
@@ -143,7 +144,7 @@ export class TransactionCommand {
             if (this.mode === TransactionCommandMode.AGGREGATE) {
                 const aggregate = this.calculateSuggestedMaxFee(
                     AggregateTransaction.createComplete(
-                        Deadline.create(this.epochAdjustment),
+                        this.createDeadline(this.epochAdjustment),
                         this.stageTransactions.map((t) => t.toAggregate(currentSigner)),
                         this.networkType,
                         [],
@@ -222,5 +223,13 @@ export class TransactionCommand {
             return fees || this.networkConfiguration.defaultDynamicFeeMultiplier;
         }
         return undefined;
+    }
+    protected createDeadline(deadlineInHours = 2): Deadline {
+        const deadline = Deadline.create(this.epochAdjustment, deadlineInHours);
+        const clientServerTimeDifference = AppStore.getters['network/clientServerTimeDifference'];
+        if (clientServerTimeDifference >= 0) {
+            return Deadline.createFromAdjustedValue(deadline.adjustedValue + clientServerTimeDifference);
+        }
+        return Deadline.createFromAdjustedValue(deadline.adjustedValue - clientServerTimeDifference);
     }
 }
--- a/src/services/TransactionCommand.ts
+++ b/src/services/TransactionCommand.ts
@@ -159,19 +159,13 @@ export class TransactionCommand {
                 });
 
                 const aggregate = this.calculateSuggestedMaxFee(
-                    AggregateTransaction.createBonded(
-                        Deadline.create(this.epochAdjustment, 48),
-                        signedInnerTransactions,
-                        this.networkType,
-                        [],
-                        maxFee,
-                    ),
+                    AggregateTransaction.createBonded(this.createDeadline(48), signedInnerTransactions, this.networkType, [], maxFee),
                 );
                 return account.signTransaction(aggregate, this.generationHash).pipe(
                     map((signedAggregateTransaction) => {
                         const hashLock = this.calculateSuggestedMaxFee(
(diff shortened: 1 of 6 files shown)


src/services/TransactionCommand.ts
```diff
@@ -223,4 +224,12 @@ export class TransactionCommand {
         }
         return undefined;
     }
+    protected createDeadline(deadlineInHours = 2): Deadline {
+        const deadline = Deadline.create(this.epochAdjustment, deadlineInHours);
+        const clientServerTimeDifference = AppStore.getters['network/clientServerTimeDifference'];
+        if (clientServerTimeDifference >= 0) {
+            return Deadline.createFromAdjustedValue(deadline.adjustedValue + clientServerTimeDifference);
+        }
+        return Deadline.createFromAdjustedValue(deadline.adjustedValue - clientServerTimeDifference);
```
yilmazbahadir, line 233: I think we don't need this check, just return it as below? If `clientServerTimeDifference`is negative it'll be a subtraction if positive then addition, right?

```
return Deadline.createFromAdjustedValue(deadline.adjustedValue + clientServerTimeDifference);
```

src/views/modals/ModalTransactionConfirmation/ModalTransactionConfirmationTs.ts
```diff
@@ -948,4 +950,11 @@ export class ModalTransactionConfirmationTs extends Vue {
         // // pre-store selected harvesting node in local
         this.saveHarvestingNode(accountAddress, this.command.formItems.nodeModel);
     }
+    protected createDeadline(deadlineInHours = 2): Deadline {
+        const deadline = Deadline.create(this.epochAdjustment, deadlineInHours);
+        if (this.clientServerTimeDifference >= 0) {
+            return Deadline.createFromAdjustedValue(deadline.adjustedValue + this.clientServerTimeDifference);
+        }
+        return Deadline.createFromAdjustedValue(deadline.adjustedValue - this.clientServerTimeDifference);
+    }
```
yilmazbahadir, line 959: same

src/store/Network.ts
```diff
@@ -538,6 +547,14 @@ export default {
             const peerNodes: NodeInfo[] = await nodeRepository.getNodePeers().toPromise();
             commit('peerNodes', _.uniqBy(peerNodes, 'host'));
         },
+        async SET_CLIENT_SERVER_TIME_DIFFERENCE({ getters, commit }) {
+            const repositoryFactory: RepositoryFactory = getters['repositoryFactory'] as RepositoryFactory;
```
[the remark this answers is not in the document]
yilmazbahadir: I think, in terms of efficiency, I'd probably go as you described. But I can't say his method is incorrect, I guess he preferred to create a couple of Deadline objects to calculate time offset, the downside is, it's adding an overhead.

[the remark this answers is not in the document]
cryptoBeliever: That's why I thought in this it will be easier to understand. Just proposition anyway :+1:

src/views/forms/FormTransactionBase/FormTransactionBase.ts
```diff
@@ -177,14 +180,19 @@ export class FormTransactionBase extends Vue {
...
     }
 
     /**
      * it creates the deadlines for the transactions.
      */
-    protected createDeadline(): Deadline {
-        return Deadline.create(this.epochAdjustment);
+    protected createDeadline(deadlineInHours = 2): Deadline {
+        const deadline = Deadline.create(this.epochAdjustment, deadlineInHours);
+        if (this.clientServerTimeDifference >= 0) {
```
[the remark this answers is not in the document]
yilmazbahadir: going to fix this one @bassemmagdy ?

yilmazbahadir (approved):

cryptoBeliever:
@bassemmagdy there is something wrong with the deadline for aggregate complete when creating mosaics:
https://share.getcloudapp.com/Qwu51glk

src/services/TransactionCommand.ts
```diff
@@ -143,7 +144,7 @@ export class TransactionCommand {
             if (this.mode === TransactionCommandMode.AGGREGATE) {
                 const aggregate = this.calculateSuggestedMaxFee(
                     AggregateTransaction.createComplete(
-                        Deadline.create(this.epochAdjustment),
+                        this.createDeadline(this.epochAdjustment),
```
cryptoBeliever, line 147: This line is wrong.

cryptoBeliever: Should be this.createDeadline()

cryptoBeliever:
Offline transactions now crashing with error from OfflineRepositoryFactory:

```
Requesting offline epoch adjustment and it's not implemented 
```

and next:

```
Uncaught (in promise) TypeError: Cannot read property 'toPromise' of undefined
    at Function.eval (DeadlineService.js?4f5e:85)
    at Generator.next (<anonymous>)
    at eval (DeadlineService.js?4f5e:23)
    at new Promise (<anonymous>)
    at __awaiter (DeadlineService.js?4f5e:19)
    at Function.create (DeadlineService.js?4f5e:84)
    at _callee13$ (Network.ts?a6d1:553)
    at tryCatch (runtime.js?96cf:63)
    at Generator.invoke [as _invoke] (runtime.js?96cf:294)
    at Generator.eval [as next] (runtime.js?96cf:119)

```
video: https://share.getcloudapp.com/Z4uj8OeY

cryptoBeliever:
All issues fixed :+1:

yilmazbahadir (approved):

PR #1723, opened 2021-09-29
fix: showing a warning on transactions detail/cosignature modal

### LINKED ISSUE #1719: Additional warning for aggregate bonded announced by "third party" address
Reason:

Scammers are sending aggregate bonded requesting all funds to random addresses (transaction is initialized from a "third-party" wallet). For example:
http://explorer.symbolblockchain.io/transactions/7BE1EB498B914BDA7F17C26789711E62276280F6322A8C3F36A9E60CD878C9B6

We should add a warning (after pop up with password) - double confirmation - to ask the user if he really wants to send this transaction because it was initialized by the third-party wallet.

We should warn if account which initialized tx is not related to an account that is signing (we should not warn if another cosigner from the same multi- initialized the transaction).

There are some legit bonded initialized by "third-party" accounts like those send by All nodes or nember.art and in future we should maybe do some "whitelist" of addresses.

yilmazbahadir:
fixes #1719 
showing a warning on transactions detail/cosignature modal and getting ack when the tx is announced by 3rd party account

WIP: 
* texts need to be localized
* all the transaction types need to be tested (restriction, metadata etc)

image

https://www.loom.com/share/57b7dcfb2dd14f12b7067eaa999a57e3

### DIFF
--- a/src/language/en-US.json
+++ b/src/language/en-US.json
@@ -1126,5 +1126,8 @@
     "welcome_to_symbol": "Welcome to Symbol",
     "word": "Word",
     "ws_connection_failed": "The wallet cannot monitor the activities of your account on the Symbol chain. Please try selecting a different node.",
-    "x_seconds": "{seconds}s."
+    "x_seconds": "{seconds}s.",
+    "transaction_cosignature_warning_dont_sign": "Do not sign if you do not know the origin of the transaction.",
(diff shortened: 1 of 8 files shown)


cryptoBeliever:
I think only case we should avoid is warning for cosigners when we creating multisig: https://share.getcloudapp.com/9ZuQZnWK

cryptoBeliever (approved):

PR #1733, opened 2021-10-08
fix: multilevel multisig account creation issue fixed

yilmazbahadir:


cryptoBeliever:
@yilmazbahadir now I'm able to sign transactions (password input is visible). Could we also fix the tick presented in the Height column on the list? It should be presented only for signed transactions. 

image

PR #1769, opened 2021-11-15
fix #1767

### LINKED ISSUE #1767: Error login - when selected node websockets doesn't work
Looks like it's issue with nodes that don't support webscockets or websockets doesn't work. 

Steps to reproduce:
1. Create profile and connect to node X
2. Close app
3. Shut down node X
4. Open app and login
5. App will detect that node X is down and will switch to other node Y

If node Y will be without working WebSockets user will be not able to login (error below) each time.

Steps to reproduce (easier way for developer):
1. Create profile on mainnet (can be on browser version because it's easier to change local storage data)
2. Open chrome console go to Application -> Local storage -> profiles. Change for profile selectedNodeUrlToConnect to (for example) https://0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-4.symbol-nodes.jp:3001/
3. Close chrome console and try login. Each time failed (for 0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-0-4.symbol-nodes.jp node WebSockets doesn't work).

image

**Workaround for users**:
- **If user have mnemonics (24 words) backup**: Try create a new profile using mnemonics (24 words) backup. Create a new profile -> Import mnemonics. It should connect to random node which hopefully will have working websockets.
- **If user don't have mnemonics backup**. User should wait until the new version released. 

Also if node has disabled wss web sockets but enabled HTTPS user is not able switch to this node:

image

bassemmagdy:
Currently, when a node allows SSL/port:3001 and does not provide a stable connection for WebSockets subscription the wallet fails to login.

Proposed solution:
fallback for HTTP/port:3000 and retry the connection, in case it fails wallet should fall-back to a different node seeking a valid Websocets connection.

### DIFF
--- a/src/services/NetworkService.ts
+++ b/src/services/NetworkService.ts
@@ -25,6 +25,7 @@ import { combineLatest, defer, EMPTY, from, Observable } from 'rxjs';
 import { catchError, flatMap, map, tap } from 'rxjs/operators';
 import { NetworkConfiguration, NetworkType, RepositoryFactory, RepositoryFactoryHttp } from 'symbol-sdk';
 import { OfflineRepositoryFactory } from '@/services/offline/OfflineRepositoryFactory';
+import { CommonHelpers } from '@/core/utils/CommonHelpers';
 
 /**
  * The service in charge of loading and caching anything related to Network from Rest.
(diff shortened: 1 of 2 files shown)


bassemmagdy:
rebased in https://github.com/symbol/desktop-wallet/pull/1770

PR #1770, opened 2021-11-15
fix #1767

### LINKED ISSUE #1767: Error login - when selected node websockets doesn't work

bassemmagdy:
Currently, when a node allows SSL/port:3001 and does not provide a stable connection for WebSockets subscription the wallet fails to login.

Proposed solution:
fallback for HTTP/port:3000 and retry the connection, in case it fails wallet should fall-back to a different node seeking a valid Websocets connection.

### DIFF
--- a/src/services/NetworkService.ts
+++ b/src/services/NetworkService.ts
@@ -25,7 +25,7 @@ import { combineLatest, defer, EMPTY, from, Observable } from 'rxjs';
 import { catchError, flatMap, map, tap } from 'rxjs/operators';
 import { NetworkConfiguration, NetworkType, RepositoryFactory, RepositoryFactoryHttp } from 'symbol-sdk';
 import { OfflineRepositoryFactory } from '@/services/offline/OfflineRepositoryFactory';
-
+import { CommonHelpers } from '@/core/utils/CommonHelpers';
 /**
  * The service in charge of loading and caching anything related to Network from Rest.
  * The cache is done by storing the payloads in SimpleObjectStorage.
@@ -56,11 +56,12 @@ export class NetworkService {
         nodeUrl: string,
         defaultNetworkType: NetworkType = NetworkType.TEST_NET,
         isOffline = false,
+        nodeWsUrl?: string,
     ): Observable<{
         networkModel: NetworkModel;
         repositoryFactory: RepositoryFactory;
     }> {
-        return from(this.createRepositoryFactory(nodeUrl, isOffline, defaultNetworkType)).pipe(
+        return from(this.createRepositoryFactory(nodeUrl, isOffline, defaultNetworkType, nodeWsUrl)).pipe(
             flatMap(({ url, repositoryFactory }) => {
                 return repositoryFactory.getGenerationHash().pipe(
                     flatMap((generationHash) => {
@@ -110,9 +111,10 @@ export class NetworkService {
         url: string,
         isOffline = false,
         networkType = NetworkType.TEST_NET,
+        nodeWsUrl?: string,
     ): Observable<{ url: string; repositoryFactory: RepositoryFactory }> {
         // console.log(`Testing ${url}`);
-        const repositoryFactory = NetworkService.createRepositoryFactory(url, isOffline, networkType);
+        const repositoryFactory = NetworkService.createRepositoryFactory(url, isOffline, networkType, nodeWsUrl);
         return defer(() => {
             return repositoryFactory.getGenerationHash();
         }).pipe(
@@ -155,16 +157,35 @@ export class NetworkService {
     public reset(generationHash: string) {
         this.storage.remove(generationHash);
     }
+    /**
+     * It checks if a node has Websocket functioning properly to subscribe.
+     * @param url the url.
+     * @param timeout number
+     */
+    public async checkWebsocketConnection(url: string, timeout: number): Promise<boolean> {
+        const webSocket = new WebSocket(url);
+        let websocketConnectionStatus: boolean = false;
+        webSocket.onmessage = function (e: any) {
+            websocketConnectionStatus = e.currentTarget.readyState == 1;
+        };
+        await CommonHelpers.sleep(timeout);
+        return websocketConnectionStatus;
+    }
 
     /**
      * It creates the RepositoryFactory used to build the http repository/clients and listeners.
      * @param url the url.
      */
-    public static createRepositoryFactory(url: string, isOffline: boolean = false, networkType = NetworkType.TEST_NET): RepositoryFactory {
+    public static createRepositoryFactory(
+        url: string,
+        isOffline: boolean = false,
+        networkType = NetworkType.TEST_NET,
+        nodeWsUrl?: string,
+    ): RepositoryFactory {
         return isOffline
             ? new OfflineRepositoryFactory(networkType)
             : new RepositoryFactoryHttp(url, {
-                  websocketUrl: URLHelpers.httpToWsUrl(url) + '/ws',
+                  websocketUrl: nodeWsUrl || URLHelpers.httpToWsUrl(url) + '/ws',
                   websocketInjected: WebSocket,
               });
     }
--- a/src/store/Network.ts
+++ b/src/store/Network.ts
@@ -309,11 +309,28 @@ export default {
                     progressCurrentNodeIndex: 1,
                     progressTotalNumOfNodes: 1,
                 });
-                nodeNetworkModelResult = await networkService.getNetworkModel(newCandidateUrl, networkType, isOffline).toPromise();
+                const nodeService = new NodeService();
+                const statisticsServiceNodes = !isOffline ? await nodeService.getNodesFromStatisticService(networkType) : undefined;
+                const nodesList = statisticsServiceNodes || nodeService.loadNodes(currentProfile);
+
+                const nodeWsUrl = nodesList.find((n) => n.url === newCandidateUrl)?.wsUrl;
+                nodeNetworkModelResult = await networkService
+                    .getNetworkModel(newCandidateUrl, networkType, isOffline, nodeWsUrl)
+                    .toPromise();
+                let websocketConnectionStatus = false;
+                if (nodeNetworkModelResult && navigator.onLine && !isOffline) {
+                    websocketConnectionStatus = await networkService.checkWebsocketConnection(
+                        nodeNetworkModelResult.repositoryFactory.websocketUrl,
+                        2000,
+                    );
+                }
                 if (
                     nodeNetworkModelResult &&
+                    nodeNetworkModelResult.repositoryFactory &&
+                    nodeNetworkModelResult.repositoryFactory.websocketUrl &&
                     nodeNetworkModelResult.networkModel &&
-                    nodeNetworkModelResult.networkModel.networkType === networkType
+                    nodeNetworkModelResult.networkModel.networkType === networkType &&
+                    websocketConnectionStatus
                 ) {
                     await dispatch('CONNECT_TO_A_VALID_NODE', nodeNetworkModelResult);
                 } else {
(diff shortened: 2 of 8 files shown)


src/services/NetworkService.ts
```diff
@@ -168,4 +169,18 @@ export class NetworkService {
...
     }
+    /**
+     * It checks if a node has Websocket functioning properly to subscribe.
+     * @param url the url.
+     */
+    public async checkWebsocketConnection(url: string) {
+        const webSocket = new WebSocket(url);
+        let websocketConnectionStatus: boolean = false;
+        webSocket.onmessage = function (e) {
+            // @ts-ignore
```
OlegMakarenko, line 180: Can we avoid `// @ts-ignore`?

src/services/NetworkService.ts
```diff
@@ -168,4 +169,18 @@ export class NetworkService {
...
+            websocketConnectionStatus = e.currentTarget.readyState == 1;
+        };
+        await CommonHelpers.sleep(2000);
```
OlegMakarenko, line 183: Why do we need this delay?

src/store/Network.ts
```diff
@@ -311,10 +311,25 @@ export default {
...
                 });
                 nodeNetworkModelResult = await networkService.getNetworkModel(newCandidateUrl, networkType, isOffline).toPromise();
+                let wsUrl = undefined;
+                let wsConnection = undefined;
+                if (navigator.onLine && !isOffline) {
+                    wsUrl = nodeNetworkModelResult.repositoryFactory.createListener().url;
+                    wsConnection = await networkService.checkWebsocketConnection(wsUrl);
+                    if (!wsConnection.status && wsConnection.sslNode) {
+                        nodeNetworkModelResult = await networkService
+                            .getNetworkModel(newCandidateUrl.replace('https', 'http').replace('3001', '3000'), networkType, isOffline)
```
OlegMakarenko, line 321: I think ports should not be hardcoded and returned by statistics-service

bassemmagdy: this is for the fallback case, sometimes the node allows WS connection on  port 3000 but not 3001, while the node itself is providing ssl connection, so we need to replace the returned one and retry.

src/services/NetworkService.ts
bassemmagdy: to make sure readyState has the correct value returned from the function.

src/services/NetworkService.ts
bassemmagdy: it's needed here to skip the linter check.

Screen Shot 2021-11-15 at 6 11 08 PM

yilmazbahadir: maybe arrow function with the parameter type(at least `any`)?

src/services/NetworkService.ts
```diff
@@ -168,4 +169,18 @@ export class NetworkService {
                   websocketInjected: WebSocket,
               });
     }
+    /**
+     * It checks if a node has Websocket functioning properly to subscribe.
+     * @param url the url.
+     */
+    public async checkWebsocketConnection(url: string) {
```
yilmazbahadir, line 176: can we add a return type?

src/services/NetworkService.ts
yilmazbahadir: Can we replace `2000` with a timeout parameter here? (which is defaulted to `2000`)

src/store/Network.ts
```diff
@@ -340,10 +355,29 @@ export default {
...
+                                .getNetworkModel(
+                                    currentProfile.selectedNodeUrlToConnect.replace('https', 'http').replace('3001', '3000'),
+                                    networkType,
+                                    isOffline,
+                                )
+                                .toPromise();
+                            wsUrl = nodeNetworkModelResult.repositoryFactory.createListener().url;
+                            wsConnection = await networkService.checkWebsocketConnection(wsUrl);
+                        }
+                    }
```
yilmazbahadir, line 374: This code block seems to be duplicated. Is it possible to extract this block to an inline arrow function that returns a tuple and use both here(`else` part) and the `if` part?

src/services/NetworkService.ts
```diff
@@ -168,4 +169,18 @@ export class NetworkService {
...
+        const webSocket = new WebSocket(url);
+        let websocketConnectionStatus: boolean = false;
+        webSocket.onmessage = function (e) {
+            // @ts-ignore
+            websocketConnectionStatus = e.currentTarget.readyState == 1;
+        };
+        await CommonHelpers.sleep(2000);
+        return { status: websocketConnectionStatus, sslNode: url.indexOf(':3001') !== -1 };
```
yilmazbahadir, line 184: Would be nice to extract `3000` and `3001` to constants and all references to them.

AnthonyLaw:
I had created the PR related to the WebSocket issue.
https://github.com/symbol/statistics-service/pull/108

Should we wait for this merge then only update the wallet apps?

src/store/Network.ts
bassemmagdy: I did refactor this applying the changes in network service.

src/services/NetworkService.ts
bassemmagdy: Added

cryptoBeliever:
I tested following scenario (which I think works on 1.0.8)

1. Choose node X in profile Y
2. Close app
3. Stop node X
4. Open app and try login to profile Y

Result: Login is not possible (same error as previously)

===

Other issue is that if we have "broken" node and it will be returned by statistics service then we are also unable to login because /network/fees/transaction failing. You can simulate by setting node https://node.finefettle.life:3001/network in local storage as selected node.
Currently this node is broken. Call https://node.finefettle.life:3001/network/fees/transaction returns error 500:

```
https://node.finefettle.life:3001/network/fees/transaction

```
{
"code": "Internal",
"message": "ENOENT: no such file or directory, open '/symbol-workdir/api-node-config/config-node.properties'"
}
```

**Comment by bassemmagdy** [comment]
@cryptoBeliever 

>  I'm wonder what will happen if REST is available only over https (no http)? And no wss:3001 or ws:3000 is available. It will switch to another node?

yes, it should switch to a different node.

for the second scenario if endpoint` network/fees/transaction` returns `500`, if we stopped calling that endpoint on login/import accounts the case when a user has any transactions that need rentalFees like namespaces/mosaics in history we won't be able to display them correctly.

**Review by NikolaiB — DISMISSED** [review]

**Comment by cryptoBeliever** [comment]
> for the second scenario if endpoint `network/fees/transaction` returns `500`, actually we need to call that endpoint on login to be able to display namespaces/mosaics transactions as the wallet needs rental fees to display them correctly.

@bassemmagdy I'm not asking to not call this endpoint but broken node should not block user from login.

File: src/services/NetworkService.ts [hunk 7]
```diff
@@ -168,4 +171,51 @@ export class NetworkService {
...
+    }
+
+    /**
+     * It checks if a node has Websocket functioning properly to subscribe.
+     * @param nodeNetworkModelResult any
+     * @param url the url.
+     * @param networkType
+     * @param isOffline whether user is in offline mode
+     */
+    public async CheckWebSocketConnectionHttpAndHttps(
```
yilmazbahadir, line 198: starts with uppercase?

bassemmagdy:
@yilmazbahadir after discussion with @cryptoBeliever we agreed to separate the rental Fees [issue](https://github.com/symbol/desktop-wallet/issues/1773) into a new one and to be addressed in a different branch.

yilmazbahadir:
There 2 changes recently made in the statistics service: 

1. NodeInfo.apiStatus.webSocket object https://github.com/symbol/statistics-service/blob/8ea683a3e75591f6de9b04f7f8c106edda36466a/openapi/openapi.yml#L315
2. Filtering secured(HTTPS && WSS) nodes only.

PR https://github.com/symbol/desktop-wallet/pull/1777 integrates statistics-service openapi client with these changes.

We should be using `apiStatus.webSocket.url` instead of falling back to a different websocket protocol logic here.

You will find a new field `wsUrl` in the `NodeModel` class. We should be using that field to connect to a websocket, if not successful then try a different node.

yilmazbahadir (commented):
Where do we try to connect to the returned `wsUrl` and check it is not failing?

yilmazbahadir:
We should be using `nodeModel.wsUrl` instead of trying to calculate ourselves here => https://github.com/bassemmagdy/symbol-desktop-wallet/blob/10381d9fd7d16b5b2d4760f53787604cdcc3ca4d/src/services/NetworkService.ts#L181

bassemmagdy:
@yilmazbahadir the issue is it's not available (nodeMode.wsUrl) before creating the nodeModel, which has the createRepositroyFactory() as a part of it, so we could update the HTTPS nodes the same way as we do for HTTP ones.

src/services/NetworkService.ts
```diff
@@ -164,7 +178,7 @@ export class NetworkService {
         return isOffline
             ? new OfflineRepositoryFactory(networkType)
             : new RepositoryFactoryHttp(url, {
-                  websocketUrl: URLHelpers.httpToWsUrl(url) + '/ws',
+                  websocketUrl: url.indexOf('http:') !== -1 ? URLHelpers.httpToWsUrl(url) + '/ws' : URLHelpers.httpsToWsUrl(url) + '/wss',
```
yilmazbahadir, line 181: Do we really need this and also `httpsToWsUrl`?
It looks like `httpToWsUrl` already replaces `http` with `ws` thus if the protocol is "https" when it is replaced with "ws" it will become "ws**s**", no? And trailer `/wss` looks wrong.

yilmazbahadir:
> @yilmazbahadir the issue is it's not available (nodeMode.wsUrl) before creating the nodeModel, which has the createRepositroyFactory() as a part of it, so we could update the HTTPS nodes the same way as we do for HTTP ones.

Should we change the order of calls instead? 
First call `NodeService.getNodesFromStatisticService` and then the others?

src/views/forms/FormNodeEdit/FormNodeEditTs.ts
```diff
@@ -100,6 +100,9 @@ export class FormNodeEditTs extends Vue {
             }
             this.formItems.networkType = NetworkType[networkModel.networkType];
             this.formItems.networkHash = networkModel.generationHash;
+        } catch (err) {
+            console.log(err);
```
AnthonyLaw, line 104: We can use  rather then log `console.error(err)`

yilmazbahadir:
Looks good to me, can be merged after the checks are successful.

yilmazbahadir (dismissed):

cryptoBeliever:
@bassemmagdy 

The following scenario doesn't work again:

1. Choose node X in profile Y
2. Close app
3. Stop node X
4. Open app and try login to profile Y

Result: Login is not possible (same error as previously)

You can simulate node down by manually changing selected node in local storage to non-existent.

cryptoBeliever:
After latest fixes all good :+1:

cryptoBeliever (approved):

PR #1775, opened 2021-11-17
replace old explorer url with the new one and remove edit option

### LINKED ISSUE #1766: Explorer - URL tied to profile
Currently, when we change the URL for explorer for old profiles old URL stays.

Why it's saved on profile level? To allow edit it? @coiki do you think anyone is using this option?

My propositions:
- remove the possibility to edit explore link **or**
- provide versioning. If the version increases then overwrite URL stored on profile level with current one.

image

@coiki what you think about that?

bassemmagdy:
-fixes  #1766


src/core/database/storage/SettingsModelStorage.ts
```diff
@@ -84,6 +84,10 @@ export class SettingsModelStorage extends VersionedObjectStorage<Record<string,
                     description: 'Reset for fees update.',
                     migrate: () => undefined,
                 },
+                {
```
cryptoBeliever, line 87: - We need reset storage to remove explorer link from there?
- description is wrong

src/views/forms/FormGeneralSettings/FormGeneralSettings.vue
```diff
@@ -42,6 +40,15 @@
                         </template>
                     </FormRow>
 
+                    <FormRow>
+                        <template v-slot:label> {{ $t('set_explorer_link') }}: </template>
```
OlegMakarenko, line 44: Maybe for consistency with the labels above, we should change it to this:
```suggestion
                        <template v-slot:label> {{ $t('form_label_explorer_link') }}: </template>
```
Then the translation key should be also updated

bassemmagdy: Updated.

cryptoBeliever:
Tested. Works well.

cryptoBeliever (approved):

bassemmagdy:
rebased in https://github.com/symbol/desktop-wallet/pull/1784

PR #1781, opened 2021-11-22
harvesting page improvements, nodePublicKey and multi-device support

### LINKED ISSUE #1762: Harvesting - improvements using statistics service - query node by nodePublicKey
Blocking issue: https://github.com/symbol/statistics-service/issues/95

Using statistics service we can improve the harvesting view:

- query node on which user started harvesting directly from statistics service instead of saving in local storage. For that in state service will be prepared option to query node by nodePublicKey (https://github.com/symbol/statistics-service/issues/95). It will be possible to directly use the key from the node key link to check if an account is harvesting currently on that node.

- list of nodes can be also fetched from statistics service instead /node/peers. This probably requires pagination (https://github.com/symbol/statistics-service/issues/74) so we can split into separate task if needed.

yilmazbahadir:
fixes #1762 
* find node by nodePublicKey from stats service
* multi-device operation support

### DIFF
--- a/src/services/NodeService.ts
+++ b/src/services/NodeService.ts
@@ -111,19 +111,19 @@ export class NodeService {
     ): Promise<NodeModel> {
         try {
             const nodeInfo = await nodeGetter(paramValue);
-            if (!nodeInfo) {
-                return undefined;
+            if (nodeInfo) {
+                return this.createNodeModel(
+                    nodeInfo.apiStatus?.restGatewayUrl,
+                    networkType,
+                    nodeInfo.friendlyName,
+                    true,
+                    nodeInfo.publicKey,
+                    nodeInfo.apiStatus?.nodePublicKey,
+                    nodeInfo.apiStatus?.webSocket?.url,
+                );
             }
-            return this.createNodeModel(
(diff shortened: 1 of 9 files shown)


src/services/NodeService.ts
```diff
@@ -111,19 +111,19 @@ export class NodeService {
...
-                nodeInfo.apiStatus?.restGatewayUrl,
-                networkType,
-                nodeInfo.friendlyName,
-                true,
-                nodeInfo.publicKey,
-                nodeInfo.apiStatus?.nodePublicKey,
-                nodeInfo.apiStatus?.webSocket?.url,
-            );
         } catch (error) {
+            console.log(error);
```
rg911, line 126: no need to send to the form?

yilmazbahadir: Thanks for the comment mate, what do you mean by _form_?

rg911: I mean if it errors out, should the frontend show sth in the notification popup?

rg911 (approved):

cryptoBeliever (approved):

PR #1783, opened 2021-11-24
show update warning when user is using older versions 

### LINKED ISSUE #1393: Notification about new version in wallet
The wallet should give information to the user that a new version is available.
If possible, we should also consider the "Update" button that will allow auto-update.

bassemmagdy:
fixes #1393


src/services/VersionCheckerService.ts
```diff
@@ -0,0 +1,52 @@
...
+ */
+import { appConfig } from '@/config';
+import fetch from 'node-fetch';
+
+/**
+ * Request github api for latest release
+ * @internal
+ * @return {Promise<LatestVersionObject>}
+ */
+const request = async (): Promise<LatestVersionObject> => {
```
OlegMakarenko, line 24: I would move it to a private method of the class below

src/services/VersionCheckerService.ts
```diff
@@ -0,0 +1,52 @@
...
+    let versionTag;
+    let downloadUrl;
+    await fetch(appConfig.repositoryDataUrl)
+        .then((response) => response.json()) //Converting the response to a JSON object
+        .then((data) => {
+            versionTag = data.tag_name;
+            downloadUrl = data.html_url;
+        })
+        .catch((error) => console.error(error));
+    return { versionTag, downloadUrl };
```
OlegMakarenko, line 34: The function should return `Promise<{versionTag: string, downloadUrl: string}` 
but in case of error it will be `Promise<{versionTag: undefined, downloadUrl: undefined}`.

Might it be better to let the error propagation and remove `.catch`?

src/services/VersionCheckerService.ts
```diff
@@ -0,0 +1,52 @@
...
+export class VersionCheckerService {
+    /**
+     * Get latest version tag
+     * @return {Promise<{versionTag:string, downloadUrl:string}>}
+     */
+    public async getLatestVersionTag(): Promise<LatestVersionObject> {
+        const versionObject = await request();
+        const versionTag = versionObject.versionTag.substring(1);
+        const downloadUrl = versionObject.downloadUrl;
+        return { versionTag, downloadUrl };
```
OlegMakarenko, line 50: What do you think if we move the app version check here?

src/services/VersionCheckerService.ts
bassemmagdy: I'd better leave the catch to see what's going on if something and handle the other case of `versionTag` might be undefined.

src/views/pages/profiles/LoginPage.vue
```diff
@@ -9,6 +9,12 @@
                 <img class="language_icon" :src="require('@/views/resources/img/login/language.svg')" alt="" />
                 <LanguageSelector />
             </div>
+
+            <div v-if="versionCheckerObject && !latestVersionInUse" class="update-box" role="alert">
+                <b syle="">{{ $t('new_version_available') }}</b>
```
OlegMakarenko, line 14: Can you please resolve this issue from Sonar?

> Replace this `<b>` tag by `<strong>`.

```suggestion
                <strong>{{ $t('new_version_available') }}</strong>
```

src/views/pages/profiles/LoginPage.less
```diff
@@ -376,6 +376,20 @@
...
+    text-align: center;
+    height: 0.3rem;
+    width: 100%;
+    display: inline-block;
+    font-size: larger;
+    line-height: 0.3rem;
+}
+
+.download-text-color {
+    color: #44004e;
```
OlegMakarenko, line 391:
```suggestion
    color: @primary;
```

cryptoBeliever:
@bassemmagdy on standalone app and Chromium "Download now" is not clickable for me. In Firefox it works.

cryptoBeliever:
Problem from https://github.com/symbol/desktop-wallet/pull/1783#issuecomment-979823593 fixed.

cryptoBeliever (approved):

src/services/VersionCheckerService.ts
```diff
@@ -0,0 +1,59 @@
...
+    downloadUrl: string;
+};
+
+export class VersionCheckerService {
+    /**
+     * Get latest version tag object
+     * @return {Promise<{versionTag:string, downloadUrl:string}>}
+     */
+
+    public async getLatestVersionTag(): Promise<LatestVersionObject> {
```
yilmazbahadir, line 30: should we rename this to `getNewerVersionTag` ? 

Current name implies that it returns the latest version no matter what. But it returns undefined if the current version is the latest. Should update the method description as well.

src/services/VersionCheckerService.ts
```diff
@@ -0,0 +1,59 @@
...
+        const versionTag = versionObject.versionTag;
+        const downloadUrl = versionObject.downloadUrl;
+        return { versionTag, downloadUrl };
+    }
+
+    /**
+     * Request github api for latest release
+     * @return {Promise<LatestVersionObject>}
+     */
+    private async requestGithubRepo(): Promise<LatestVersionObject> {
```
yilmazbahadir, line 44: rename to `fetchLatestVersionFromGithubRepo`?

src/services/VersionCheckerService.ts
```diff
@@ -0,0 +1,59 @@
...
+        let versionTag;
```
yilmazbahadir, line 45: We should avoid using `let` as much as possible. Below is an example of how it might look:

```
private async fetchLatestVersionFromGithubRepo(): Promise<LatestVersionObject> {
  const [versionTag, downloadUrl] = await fetch('https://api.github.com/repos/symbol/desktop-wallet/releases/latest')
  .then((response) => response.json()) //Converting the response to a JSON object
  .then((data) => [
    data.tag_name.substring(1),
    data.html_url
  ])
  .catch((error) => console.error(error));
  if (!versionTag || !downloadUrl) {
    return;
  }
  return { versionTag, downloadUrl };
}
```

src/services/VersionCheckerService.ts
AnthonyLaw: IMO this method is kind of redundant. `requestGithubRepo` returns the same Object.
We can just use that.

OlegMakarenko:
@bassemmagdy could you please resolve conflicts

yilmazbahadir (approved):

AnthonyLaw (approved):

PR #1805, opened 2021-12-01
fix offline transactions issue

bassemmagdy:
Offline Transaction page need a node url to be able to mock transaction data

Old Behavior: - Wallet is not able to fetch nodes from statistics service and end up with one node in storage which get filtered and emptied before mocking the data.

Introduced solution: 
- Query the storage to get urls to which that profile was previously connected and use them to mock the needed data.
- In case the wallet is never connected to the internet use a fake node to mock repositories and display UI correctly.

### DIFF
--- a/src/config/AppConfig.ts
+++ b/src/config/AppConfig.ts
@@ -35,7 +35,7 @@ export interface AppConfig {
         DECIMAL_SEPARATOR: string;
     };
     title: string;
-    marketServerUrl: string;
+    offlineNodeModelUrl: string;
 }
 
 const defaultAppConfig: AppConfig = {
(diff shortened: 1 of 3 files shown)


cryptoBeliever:
Basic scenario is: Wallet is always offline. So no node URL should be required to load offline transactions view.

1. Turn off Internet.
2. Open wallet in incognito (no history)
3. Create new profile
4. Logout 
5. Go to offline transactions view.

Result: You have error:

image

When Internet is ON it started to working but main use case is when Internet is always off on computer.

NikolaiB:
@bassemmagdy @cryptoBeliever with latest update it works for me

cryptoBeliever:
> @bassemmagdy @cryptoBeliever with latest update it works for me

Same for me.

cryptoBeliever (approved):

PR #1814, opened 2021-12-09
Feat: blacklist whitelist contacts

bassemmagdy:
- Added feature blacklist/whitelist contacts
- Separated contact list to blacklisted and whitelisted
- By default now, transactions received from blacklisted contacts don't appear in transaction history


src/services/TransactionFilterService.ts
```diff
@@ -26,9 +27,13 @@ export class TransactionFilterService {
      * @param state for extracting transactions filter options and list filtered by group.
      * @param currentSignerAddress selected signer address(es)
      */
-    public static filter(state: TransactionState, currentSignerAddress: string | string[]): Transaction[] {
+    public static filter(
+        state: TransactionState,
+        currentSignerAddress: string | string[],
+        BlackListedContacts?: IContact[],
+    ): Transaction[] {
         const { filterOptions, transactions, confirmedTransactions, unconfirmedTransactions, partialTransactions } = state;
```
Jaguar0625, line 35: should we add tests for this new functionality?

src/components/ContactDetailPanel/ContactDetailPanelTs.ts
```diff
@@ -85,8 +86,25 @@ export class ContactDetailPanelTs extends Vue {
         this.showDeleteConfirmModal = val;
     }
 
+    public get showBlackWhiteListModal() {
+        return this.showBlackWhiteListConfirmModal;
+    }
+    public set showBlackWhiteListModal(val: boolean) {
```
Jaguar0625, line 92: please add newline after closing brace

src/components/ContactDetailPanel/ContactDetailPanelTs.ts
```diff
@@ -85,8 +86,25 @@ export class ContactDetailPanelTs extends Vue {
...
     public removeContact() {
         this.$store.dispatch('addressBook/REMOVE_CONTACT', this.selectedContact.id);
         this.showDeleteConfirmModal = false;
     }
+
+    public BlackListWhiteListContact() {
+        this.selectedContact.isBlackListed = !this.selectedContact?.isBlackListed;
+        this.$store.dispatch('addressBook/UPDATE_CONTACT', { id: this.selectedContact.id, contact: this.selectedContact });
+        this.showBlackWhiteListModal = false;
+    }
```
Jaguar0625, line 105: newline after closing brace

src/components/ContactDetailPanel/ContactDetailPanelTs.ts
```diff
@@ -85,8 +86,25 @@ export class ContactDetailPanelTs extends Vue {
...
+    public set showBlackWhiteListModal(val: boolean) {
+        this.showBlackWhiteListConfirmModal = val;
+    }
+
     public removeContact() {
         this.$store.dispatch('addressBook/REMOVE_CONTACT', this.selectedContact.id);
         this.showDeleteConfirmModal = false;
     }
+
+    public BlackListWhiteListContact() {
```
Jaguar0625, line 101: should we rename this to ToggleBlackListContact?

src/core/utils/CommonHelpers.ts
```diff
@@ -61,4 +61,17 @@ export class CommonHelpers {
             }
         }
     }
+    /**
+     * Truncates the string if it is too long
+     * @protected
+     * @param {string} str
+     * @returns {string}
+     */
+    public static truncate(str: string): string {
```
Jaguar0625, line 70: consider adding tests for this

src/components/TransactionList/TransactionListFilters/TransactionListFiltersTs.ts
```diff
@@ -70,4 +82,22 @@ export class TransactionListFiltersTs extends Vue {
     public downloadTransactions() {
         this.$emit('downloadTransactions');
     }
+    onSelectBlackListed() {
```
Jaguar0625, line 85: 1. newline after brace
2. if we have unit tests for this component, we should add some for the new functionality

bassemmagdy:
@Jaguar0625 All comments are addressed, please recheck :)

gimre-xymcity (dismissed):

__tests__/utils/CommonHelpers.spec.ts
```diff
@@ -32,4 +32,15 @@ describe('utils/CommonHelpers', () => {
...
     });
+
+    describe('truncate(address)', () => {
+        test('returns truncated string', () => {
+            const address = Address.createFromRawAddress('TAD5BAHLOIXCRRB6GU2H72HPXMBBVAEUQRYPHBY');
+            const truncatedAddress = CommonHelpers.truncate(address.plain());
+            const string = 'string';
+            expect(truncatedAddress.length).toBe(15);
+            expect(truncatedAddress.substring(truncatedAddress.length - 6)).toEqual('...HBY');
+            expect(CommonHelpers.truncate(string).length).toBe(6);
```
Jaguar0625, line 43: 1. it's probably better to rename `string` to something more descriptive like `shortString` 
2. in future, generally would try to group Arrange / Act / Assert [see https://hackmd.io/WrKdAaohRXKgCbzd_4ZMxA]

__tests__/services/TransactionFilterService.spec.ts
```diff
@@ -17,6 +17,8 @@ import { TransactionFilterOptionsState, TransactionState } from '@/store/Transac
 import { TransactionFilterService } from '@/services/TransactionFilterService';
 import { getTestAccount } from '@MOCKS/Accounts';
 import { TransferTransaction } from 'symbol-sdk';
+import { addressBookMock } from '@MOCKS/AddressBookMock';
+import { IContact } from 'symbol-address-book';
 
 const currentSigner = getTestAccount('remoteTestnet');
 const recepient = getTestAccount('remoteTestnetRecipient');
```
Jaguar0625, line 24: 1. not something you need to fix as part of this PR, but it's a bit confusing to use currentSigner as recipient address
2. recepient is spelled wrong

__tests__/components/TransactionList/TransactionListFilters/TransactionListFilter.spec.ts
```diff
@@ -25,8 +38,21 @@ describe('TransactionListFilters', () => {
         vm.onSignerSelectorChange('TAD5BAHLOIXCRRB6GU2H72HPXMBBVAEUQRYPHBY');
         expect(vm.$store.dispatch).toBeCalledWith('account/SET_CURRENT_SIGNER', { address: addr, reset: true, unsubscribeWS: false });
     });
+
     test("should not call the 'account/SET_CURRENT_SIGNER' without address", () => {
         vm.onSignerSelectorChange();
         expect(vm.$store.dispatch).not.toBeCalled();
     });
+
+    test("should call 'transaction/filterTransactions' with blacklisted contacts", () => {
```
Jaguar0625, line 47: would be good to add a test when `isBlackListFilterActivated == true` to cover else branch too

bassemmagdy:
@Jaguar0625 All feedbacks should be now addressed :)

cryptoBeliever:
Pasting here as a reference. When I'm doing on branch:
npm install
npm run dev

I got error like this:

image

cryptoBeliever:
@bassemmagdy @coiki I think the address book/blocked transactions mechanic and UX/UI has to be re-think. It's not intuitive for me.

1) I'm able to add mainnet addresses to testnet profile which doesn't make sense I guess. Same with testnet addresses in mainnet profile.

2) Have we graphics to address the book? Currently white and blacklist sublists are aligned to the left which looks strange. Also empty 

image
image

3) Option to import an address book is only available if you have an empty address book (white and black lists). How user should import more addresses to black/white lists? Should be possible?

4) On the transaction list "Show blacklisted transactions" button has no indicator that it's active. It's very misleading for users. Maybe besides the "active" filtering indicator, we should add some iconographic in the transaction row?

How filtering should work? It should show valid + blacklisted transactions or should show only blacklisted? Currently, it's the second one. Could you tell me how filtering currently works for aggregate transactions where signed and initiator can be different? Which field is used to filter?

https://share.getcloudapp.com/7Kuqg54L

5) First I have to add to whitelist addresses and next move it to blacklisted. There is no way to add it directly into blacklisted.

https://share.getcloudapp.com/5zunp6Z6

cryptoBeliever:
@coiki 
5) Also I think the manually managed list of white/blacklists will not protect beginners users from scams if there will be no difference between whitelisting and missing address in the book. Could you consider adding a warning (for example "This transaction is from an untrusted address" when the user is signing a transaction initialized by not whitelisted address)? 
I think it's something worth considering.

bassemmagdy:
@cryptoBeliever 
1- added checker to validate network type before adding new contact.
2- you can find the design here -> https://www.figma.com/file/lkhDQWxq5FlUrMNZYIJ2wT/Symbol-Wallet---Legacy?node-id=168%3A14032
3- Currently importing is overwriting the current list that's why it's only available when the list is empty. We can improve it by letting it add the new ones to the list but, in case the address is found in the list all contact data will be overwritten except the address.
4- I will try to add a shadow to show the difference based on state (preferably we get a design for that.)
5- Mockups didn't include that option in the contact creation form.

Jaguar0625:
@cryptoBeliever - do you think the new approach is intuitive enough or do you have more suggestions for improvements?

NikolaiB:
@cryptoBeliever how to remove an address from the blacklist? Do we have this functionality?

cryptoBeliever:
> @cryptoBeliever how to delete address from the blacklist? Do we have this functionality?
@NikolaiB by using "Delete Contact" in the address book. Same as for "White List" contacts.

@bassemmagdy 
1. It's fixed.
2. It doesn't look as on Figma Graphics.

a) Figma:

image

And wallet:

image

- whitelist and blacklist header: missing grey horizontal line, header text size is same as for "My contacts" but should be way smaller
- contact item: here also text seems too big, icon is too small comparing text; icon has also other borders color than on Figma

3. I think we should allow import and overwrite. @coiki what do you think?

4. As I can see on Figma there is a graphic for the active and inactive state:

image

5. Adding to white/blacklist. I think it requires changes. My proposition is to add a select box when adding a new contact with two options blacklist and whitelist. By default could be selected whitelist if we clicked "Add Contact" when whitelist is open or blacklist if we clicked "Add contact" when blacklist is open (but user could change that). Asking user to first add into whitelist and move to blacklist doesn't look good for me. @coiki what do you think?

6. @coiki Signing partial transactions: 
image

At this moment blacklisting change only visibility on the transaction list. But maybe we should present some warning on transaction details (if the user decides to open a transaction to sign anyway or by mistake). My proposition is to not allow sign partial transactions if the user already added an address to blacklist.

7. On the list we have now a button to show blacklisted transactions. When we activate it we can see only "blacklisted" transactions. For me, more intuitive would be present in such a situation whitelisted + blacklisted transactions (button would hide/show only blacklisted transactions) - whitelisted always visible. @coiki what do you think it would be better?

8. Blacklisting transaction rules. @bassemmagdy could you write what rules are for blacklisting transactions? Transaction is blacklisted when transaction signer is on the blacklist?

9. @bassemmagdy Figma (https://www.figma.com/file/lkhDQWxq5FlUrMNZYIJ2wT/Symbol-Wallet---Legacy?node-id=168%3A14032) contains changed flow for signing partial transactions and saving addresses (black and whitelisting) to address book which weren't implemented. I think it should be part of address book implementation.

Transaction details when opened (I think for whitelisted addresses we could move to Caution warning and skip this step):

image

When the user reject there is an option to blacklist contact:

image
image

There is also an option to "Archive". @coiki what it does? It should hide transactions? It should be implemented here or there will be a separate issue for the feature (as I understand it's an option to hide transactions without blacklisting address)?

The user can also "Accept..." and on the next screen, he can see Caution, and on the next screen provides a password to accept. At the end of process, there should be a way to add contact into "Whitelist"

image
image
image

coiki:
@cryptoBeliever I checked your questions and I’ve tried to sum up my answers. We already spoke on the chat that we need to create a new task for this, or reuse the one that already touched this subject (and delete it if not).

1. Understood. I agree it doesn’t make sense.
2. The alignment should be to the left. The design is quite far from what was in the design from Biagio.
3. Users will add contacts when interacting with transactions, but should always be able to add directly.  If I understand correctly we have only 1 address book per whitelist/blacklist? We should have the option to always import an address book and it should just append it to my existing one.

Regarding showing the blacklisted transactions alongside the whitelisted, we want to treat it as a sort of ‘spam folder’ where all the toxic txs live. I think there are use cases where it would be helpful having them alongisde. Let’s see what we can do with this idea. I would also prefer view/hide, but we wanted to manage it like this due to the malicious attacks.


Regarding the new ‘signing area’ of the transactions:
I agree with your point to not allow to sign transactions from blacklisted addresses. Let’s add it to the logic of the new task about the signing area.

Q: What is Archive? A: It just hides that view (so the user only rejects but doesn’t blacklist). Let’s change the text to “Close”.

cryptoBeliever:
Reported a few issues:

I think those should be fixed in this PR:
https://github.com/symbol/desktop-wallet/issues/1823
https://github.com/symbol/desktop-wallet/issues/1824
https://github.com/symbol/desktop-wallet/issues/1825

Those could be done in separate PR:
https://github.com/symbol/desktop-wallet/issues/1826
https://github.com/symbol/desktop-wallet/issues/1827

Also depends on @coiki answers for questions in https://www.figma.com/file/lkhDQWxq5FlUrMNZYIJ2wT/Symbol-Wallet---Legacy?node-id=168%3A14032 I may report some further issues because currently there is no difference in flow between the situation when accounts is whitelisted/unknown/blacklisted.

gimre-xymcity (approved):

Issue #1843, opened 2022-02-14
Transaction details - address should be clickable and show account details in explorer


cryptoBeliever:
Current state:
Currently, in transaction details, all addresses are not clickable. But it would be good for the user to allow click it (same as transaction hash) to see account details in the explorer. Same for a contact name for resolved addresses.

![image](https://user-images.githubusercontent.com/17684281/153929803-a29a394a-e89d-4a3f-b418-54bbd8510535.png)

cryptoBeliever:
@OlegMakarenko I thought about the same thing as we have for transaction hash (just link). You did in this way to allow show address even if it's replaced by contact name?

cryptoBeliever:
The current solution looks good :+1: We have the option to open in explorer and copy the address.

![image](https://user-images.githubusercontent.com/17684281/154860045-93f3225d-b439-49d3-9d25-8bc6f7ccc2c6.png)

Fixed in https://github.com/symbol/desktop-wallet/pull/1847

cryptoBeliever:
Released in 1.0.11

Issue #1842, opened 2022-02-14
[AddressBook] - transaction detail "From" is not replaced with contact name


cryptoBeliever:
Steps:
1. We have two accounts X and Y.
2. Send transaction from X to Y.
3. Go to account Y and add account X to the address book.
4. Go to transaction history and open details for transaction from X

![image](https://user-images.githubusercontent.com/17684281/153929446-64f60894-5895-4320-b1ee-b93f17c6e975.png)

Result: In transaction details for "From" we can see address instead of the contact name.
Expected result: Show contact names for addresses that are in the address book.

cryptoBeliever:
Fixed in https://github.com/symbol/desktop-wallet/pull/1847

cryptoBeliever:
Released in 1.0.11

Issue #1841, opened 2022-02-14
Remove all dashes from addresses


cryptoBeliever:
We want to adopt standards that make Symbol easier to use. We have seen that the dashes in the addresses don’t really help and people don’t really understand them. It also adds complexity when creating a Symbol app.

AC1. Remove dashes from all addresses all across the app.

cryptoBeliever:
Done in https://github.com/symbol/desktop-wallet/pull/1847

cryptoBeliever:
Done in https://github.com/symbol/desktop-wallet/pull/1847. Will close after merge to dev.

cryptoBeliever:
Released in 1.0.11

Issue #1840, opened 2022-02-14
Present "Signer" as address, not public key


cryptoBeliever:
Currently, "Signer" in transaction details is presented using the public key.
I think we should present it as an address that is more clear for users.

![image](https://user-images.githubusercontent.com/17684281/153928185-58a0a0a4-5aa8-4763-a409-0ac9c4db92fb.png)

cryptoBeliever:
Done in https://github.com/symbol/desktop-wallet/pull/1847. Will close after merge.

cryptoBeliever:
Released in 1.0.11

PR #1847, opened 2022-02-16
fix: address book

OlegMakarenko:
**What the current behaviour is**
PR #1814 has 12 unresolved related issues.

**Why this is an issue.**
Issues must be resolved before merge.

**What's the fix**
Resolved: #1823
- Fixed styles of the whitelist/blacklist tabs.
- Fixed styles of the contact selector.
- Password controls moved to the next step.
- Fixed the checkbox artifact.
- Fixed the "Clicking in "I understand" should check the checkbox too".
- Fixed styles of the New Address Detected modal.
- Renamed "Blacklist" buttons to "Block"

Resolved: #1826 
- Added option to block new contact.

Resolved: #1834
- Fixed the "My Contacts - Not possible to edit address"

Related: #1827
- Added option to import and rewrite address book.

Resolved: #1831
- Fixed the "Aggregate transactions get random names from the addressbook"
- Fixed the issue with the tx list while new transactions added via listener.
- Fixed tx filter.

Resolved: #1739
- Fixed styles and layout of the ModalTransactionCosignature warning section (rewrote).
- Added escape options.
- Added blocked contact view.

Resolved: #1840
- Added signer address to transaction details.

Resolved: #1841
- Removed all dashes from addresses.

Resolved: #1842
- Fixed the "transaction detail "From" is not replaced with contact name"

Resolved: #1843
- Added the address popover and link to the Explorer.

Resolved: #1848
Resolved: #1849
- Fixed the "Multisig tree - presentation for multilevel multisig"

Other
- Updated translations.
- Added test.


cryptoBeliever:
@OlegMakarenko I have the following things we can improve:

1. When we open aggregate bonded transaction from unknown/known signer it scrolls down. This is something you change? I think it's better to allow users to check details from top to bottom.
2. We can blacklist also a signer who is a cosigner from our multi-sig. In such a case transaction goes to "blocked transactions" which is fine, but when I open it I got no warning about the blocked signer. I think this should be more important than the relation with multi-sig. 
3. For unknown signers we have "Transaction from unknown account requires a signature.". I would change it to something that it does not suggest signing. Maybe "Transaction sent from the unknown account, please review details and accept or reject"?
4. We have two buttons Accept/Reject. Accept has two dots ("Accept.."). Do you see a reason for that? If not I would switch both buttons to without dots.

I reported also: https://github.com/symbol/desktop-wallet/issues/1848

OlegMakarenko:
@cryptoBeliever thanks.

1. Yes I've added it for unknown address (This is default behaviour of the view, it scrolls to the sign form). Fixed, removed.
2. Fixed, show warning for blocked cosigner.
3. > please review details

We already have this message on the next view. Do you think we need to duplicate this message?

4.  I don't see any reason. Fixed, removed.

5. > I reported also: https://github.com/symbol/desktop-wallet/issues/1848 which could be fixed here or independently

Fixed.

cryptoBeliever:
@OlegMakarenko  all fixed at this moment :+1: 
Regarding 3) we agreed to change the text from "Transaction from unknown account requires signature." to "Transaction from unknown account awaiting signature."

cryptoBeliever (approved):

cryptoBeliever:
Feedback from Kristy and Orisdorch:

1) remove yellow "Caution" on white screen
2) Change warning to "You are about to sign a transaction that was not created by you. Malicious transactions can result in a total loss of funds. Please carefully review all amounts and recipient addresses, as transactions are not reversable."
3) "Added to blacklist" screen. 
- Text should be "A blacklist is a list of addressees that you do not wish to interact with. These could be associated with social engineering scams, fraud, or simply a source of spam. When an address is added to your blacklist, incoming transactions from them will be automatically blocked."
- finetune the padding of the elements here.
- Introduce new icons on the title - so the skull would go up the top, and it would be 'Added to Blacklist' (make sure to capitalize titles)
4) Blocked address view:
- Text should be "This address is on your blacklist. If you wish to accept this transaction, please remove this address from your blacklist."
- Text should be "I understand the risks." (had no full stop before)
5) Icons (refresh, CSV, skull) in transaction history are not aligned.
6) We resigning from providing blocked address name when rejecting tx. Reject > go to 'Added to Blacklist'. We should not replace address with contact name if contact name is empty. I think we should also mark blacklisted addresses with red color.
7) Transfer view - we have "Select contact" option. We should not show there blocked contacts.

cryptoBeliever:
@OlegMakarenko I think all comments from https://github.com/symbol/desktop-wallet/pull/1847#issuecomment-1050214764 are done. Translations are updated too?

src/shims/vue.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.vue' {
```
Jaguar0625, line 1: did you intend to check this file in?

src/views/modals/ModalTransactionCosignature/ModalTransactionCosignatureTs.ts
```diff
@@ -469,11 +519,52 @@ export class ModalTransactionCosignatureTs extends Vue {
...
-        const addressExists: boolean = this.addressBook.getContactByAddress(this.transaction.signer.address.plain()) !== undefined;
+    get addressExists(): boolean {
+        const addressExists = this.addressBook.getContactByAddress(this.transaction.signer.address.plain()) !== undefined;
         return addressExists;
     }
 
+    public getSignerAddressContactStatus(): 'white_list' | 'black_list' | 'unknown' {
+        const signerAddress = this.transaction.signer.address.plain();
+        const contacts = this.addressBook.getAllContacts();
+        const signerConstact = contacts.find((contact) => contact.address === signerAddress);
```
Jaguar0625, line 530: contact

Jaguar0625 (approved):

src/shims/vue.d.ts
OlegMakarenko: I've added it so TSC can recognize Vue component as a module:
https://github.com/symbol/desktop-wallet/blob/04f9ed9477cadba3e0478ccff025b886e5a7fcbc/__tests__/views/modals/ModalTransactionCosignature.spec.ts#L23

src/views/modals/ModalTransactionCosignature/ModalTransactionCosignatureTs.ts
OlegMakarenko: Thanks!

Jaguar0625 (approved):

Issue #1820, opened 2022-01-03
Harvesting: Join "Key links" and "Persistent delegated message" steps


cryptoBeliever:
At this moment harvesting in mobile and desktop wallets is working in the following way:

1. Create Account Links (Remote, VRF, Node).
2. Send a persistent delegated message to Node.

Those two steps can be joined in one aggregate transaction. This will simplify a lot of harvesting views (for single and multisig accounts).
Important is only to put a persistent delegated message in aggregate after key links transaction.

Example here:
https://symbol.fyi/transactions/43748B7F9868E624A72F8D81A394C2CBCF1A4021B5DF255501B1B9BAD2D494D6

@yilmazbahadir the question if we need graphics?

segfaultxavi:
Please update the docs if this ever gets implemented!
https://docs.symbolplatform.com/guides/harvesting/activating-delegated-harvesting-wallet.html

PR #1862, opened 2022-05-23
add one-click harvesting configuration #1820

ishidad2:
### What the current behaviour is

- Currently, harvesting in mobile and desktop wallets is working in two steps.  #1820 

### Why this is an issue

- These two transactions can be combined into one aggregate transaction

### What's the fix

1. Create Account Links (Remote, VRF, Node).
2. Send a persistent delegated message to Node.

These two transactions are combined into one and sent as an aggregate transaction.

### DIFF
--- a/src/store/Harvesting.ts
+++ b/src/store/Harvesting.ts
@@ -226,7 +226,7 @@ export default {
                     ? pollingTrials === 20 || currentSignerHarvestingModel?.delegatedHarvestingRequestFailed
                         ? HarvestingStatus.FAILED
                         : HarvestingStatus.INPROGRESS_ACTIVATION
-                    : HarvestingStatus.KEYS_LINKED;
+                    : HarvestingStatus.INACTIVE;
                 if (status === HarvestingStatus.ACTIVE && node && node[1]) {
                     // @ts-ignore
                     dispatch('UPDATE_ACCOUNT_SELECTED_HARVESTING_NODE', {
--- a/src/views/forms/FormPersistentDelegationRequestTransaction/FormPersistentDelegationRequestTransaction.vue
+++ b/src/views/forms/FormPersistentDelegationRequestTransaction/FormPersistentDelegationRequestTransaction.vue
@@ -2,7 +2,7 @@
     <div class="form-persistent-delegation-request">
         <NavigationLinks
             :direction="'horizontal'"
-            :items="['delegated_harvesting', 'key_links']"
+            :items="['delegated_harvesting']"
             :current-item-index="activePanel"
             translation-prefix="tab_harvesting_"
             @selected="(i) => (activePanel = i)"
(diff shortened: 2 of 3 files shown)


ishidad2:
I was able to reproduce it.

I had forgotten that there was a key-link-only condition in the old wallet, since the new version of the wallet does not generate a key-link-only condition.

I first reverted the "Key Link" tab back to its original state to address this issue.

Also, as another way to approach this I would consider the following
Please give me your opinion.

Issue an aggregate transaction with the unlinked transaction also included in the harvest setting.

I believe that doing the above would require a slight change in operations.
(Like Arcana wallet, when selecting a node, make an aggregate transaction that includes "LinkAction.Unlink", "LinkAction.Link", and "PersistentDelegationRequestTransaction").

In other words, the "Stop Harvesting" operation should no longer be necessary.

I believe it will take some time to achieve this, as it will require some changes in the UI.

What do you think?

cryptoBeliever:
@ishidad2 
Current logic is very complicated (FormPersistentDelegationRequestTransactionTs.ts) and needs to be analyzed in code (I think many conditions can be removed when we join those two steps) but I think we need :

1) Start harvesting:

Start harvesting should send one aggregate transaction containing:

- unlink remote account (only if an account is already linked and we don't have a private key for it in storage)
- unlink VRF key (only if an account is already linked and we don't have a private key for it in storage)
- unlink node key (if a node key is linked - we need to unlink the old one)
- link remote account (only if we had to remove the old one - then we have to generate and link a new key)
- link VRF key (only if we had to remove the old one - then we have to generate and link the new key)
- link node key (always)
- persistent delegated message (always)

2) Stop harvesting: 

Start harvesting should send one aggregate transaction containing:
- unlink node key (only if a node key is linked do we need to unlink the old one)
- unlink remote account (only if an account is already linked and we don't have the private key for it in storage)
- unlink VRF key (only if an account is already linked and we don't have the private key for it in storage)

@yilmazbahadir what do you think about it?

@ishidad2 when we decide on details you can update the issue description to provide more details of what changes introduce in this PR (here examples https://github.com/symbol/desktop-wallet/pull/1847, https://github.com/symbol/product/pull/117 - you can follow template: "What the current behaviour is", "Why this is an issue.", "What's the fix") .

ishidad2:
### What the current behaviour is

- Currently, harvesting in mobile and desktop wallets is working in two steps.  #1820 


### Why this is an issue

- These two transactions can be combined into one aggregate transaction

### What's the fix

1. Create Account Links (Remote, VRF, Node).
2. Send a persistent delegated message to Node.

These two transactions are combined into one and sent as an aggregate transaction.

yilmazbahadir:
Implemented with https://github.com/symbol/desktop-wallet/pull/1903

PR #1878, opened 2022-07-25
[wallet/desktop] task: unit tests added for Message components

OlegMakarenko:
Added tests for:
  - `MessageDisplay`
  - `MessageInput`

### DIFF
--- a/__tests__/components/MessageDisplay.spec.ts
+++ b/__tests__/components/MessageDisplay.spec.ts
@@ -1,0 +1,273 @@
+/*
+ * (C) Symbol Contributors 2022
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and limitations under the License.
+ *
+ */
+import MessageDisplay from '@/components/MessageDisplay/MessageDisplay.vue';
+import { MessageDisplayTs } from '@/components/MessageDisplay/MessageDisplayTs';
+import { NotificationType } from '@/core/utils/NotificationType';
+import { getComponent } from '@MOCKS/Components';
+import { WalletsModel1, account1 } from '@MOCKS/Accounts';
+import { Account, Address, EncryptedMessage, NamespaceId, PlainMessage, PublicAccount, UnresolvedAddress } from 'symbol-sdk';
+
+const messageDisplayTs = new MessageDisplayTs();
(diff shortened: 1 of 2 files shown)


__tests__/components/MessageDisplay.spec.ts
```diff
@@ -0,0 +1,272 @@
...
+        });
+    });
+
+    describe('onAccountUnlocked()', () => {
+        const runUnlockAccountTest = async (
+            recipient: UnresolvedAddress,
+            linkedAddress: Address | null,
+            decryptAddress: Address | null,
+            expectGetLinkedAddressToBeDispatched: boolean,
+            expectAddErrorToBeDispatched: boolean,
```
Jaguar0625, line 115: Add => Address

try to avoid abbreviations when possible

__tests__/components/MessageDisplay.spec.ts
```diff
@@ -0,0 +1,272 @@
...
+            expecDecryptMessageToBeCalled: boolean,
+        ) => {
```
Jaguar0625, line 117: consider  passing in the options via a single object. that might lead to improved readability.

__tests__/components/MessageDisplay.spec.ts
```diff
@@ -0,0 +1,272 @@
...
+        };
+
+        test('recipient is instance of NamespaceId and linkedAddress is provided', async () => {
+            // Arrange:
+            const recipient = new NamespaceId([929036875, 2226345261]);
+            const linkedAddress = account1.address;
+            const decryptAddress = linkedAddress;
+            const expectGetLinkedAddressToBeDispatched = true;
+            const expectAddErrorToBeDispatched = false;
+            const expecDecryptMessageToBeCalled = true;
```
Jaguar0625, line 174: expec*t

__tests__/components/MessageDisplay.spec.ts
Jaguar0625: given event is ADD_ERROR, either naming is fine

__tests__/components/MessageDisplay.spec.ts
Jaguar0625: i was thinking about having the calling code look something like this:

```js
            await runUnlockAccountTest(recipient, linkedAddress, decryptAddress, {
                getLinkedAddressToBeDispatched: true,
                addErrorToBeDispatched: false,
                linkedAddressToBeDispatched: true
            });
```
the first three are required and outside of the object. the latter three are part of an object (and self describing since keys are specified). in that case, the parameter can be called a more descriptive "expectations" instead of "options".

OlegMakarenko: Should we move the definition of `expectations` under `Arrange` section?

```js
// Arrange:
...
const expectations = {
    getLinkedAddressToBeDispatched: true,
    addErrorToBeDispatched: false,
    decryptMessageToBeCalled: true
}

// Act + Assert:
await runUnlockAccountTest(recipient, linkedAddress, decryptAddress, expectations);
```

Jaguar0625: i don't mind having it inline, but your choice

Jaguar0625 (approved):
👍

PR #1887, opened 2022-08-09
feature/unit tests part 6

yilmazbahadir:


__tests__/components/DurationInput.spec.ts
```diff
@@ -100,9 +102,43 @@ describe('components/DurationInput', () => {
...
         const component = wrapper.vm as DurationInputTs;
+
+        // Act:
         component.chosenValue = chosenValue;
 
         // Assert:
         expect(wrapper.emitted('input')[0][0]).toBe(chosenValue);
     });
+
+    // TODO test for validationRule
```
Jaguar0625, line 113: aren't new tests testing validationRule? is TODO still needed?

__tests__/components/DurationInput.spec.ts
```diff
@@ -100,9 +102,43 @@ describe('components/DurationInput', () => {
...
+        const wrapper = getDurationInputWrapper({ targetAsset: 'namespace', value: '10000' });
+        const component = wrapper.vm as DurationInputTs;
+
+        // Act:
+        const durationValidationRule = component.validationRule;
+
+        // Assert:
+        expect(durationValidationRule).toContain(
+            `min_value:${
+                networkConfig[NetworkType.TEST_NET].networkConfigurationDefaults.minNamespaceDuration /
```
Jaguar0625, line 139: why dividing by blockGenerationTargetTime ? is input a block count or a human readable duration ?

__tests__/components/ErrorTooltip.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+        const errors = [firstErrorItem, 'error 2'];
+        const wrapper = getErrorTooltipWrapper(errors);
+        const component = wrapper.vm as ErrorTooltipTs;
+
+        // Assert:
+        expect(component.fieldError).toBe(firstErrorItem);
+        expect(component.displayedError).toBe(firstErrorItem);
+        expect(component.errored).toBe(true);
+    });
+});
```
Jaguar0625, line 74: consider adding test where fieldError != displayedError

__tests__/components/HardwareConfirmationButton.spec.ts
```diff
@@ -0,0 +1,105 @@
...
+        jest.spyOn(TrezorConnect, 'nemSignTransaction').mockResolvedValue(expectedResult);
+
+        // Act:
+        try {
+            await component.signTransaction(transaction).toPromise();
+        } catch (error) {
+            // Assert:
+            expect(error.message).toBe(expectedResult.payload.error);
+        }
+    });
```
Jaguar0625, line 90: this and following test will pass if no exception is thrown. 
i think chai should have construction like:
expect(() => { ... }).to.throw('message or exception');

__tests__/components/FormInputEditable.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+        expect(component.value).toBe(newAccountName);
+        expect(props.onEdit).toHaveBeenCalledWith(newAccountName);
+    };
+
+    test('complete editing with success when value is not empty', async () => {
+        testEditingSuccess('myAccountName');
+    });
+
+    test('complete editing with success when value is empty', async () => {
+        testEditingSuccess();
```
Jaguar0625, line 96: can this be undefined in vue or will it always map to ''?

__tests__/components/FormInputEditable.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+        const component = wrapper.vm as FormInputEditableTs;
+
+        // Act:
+        await wrapper.find('button').trigger('click');
+        await wrapper.find('input').setValue(newAccountName);
+        component.cancelEdition();
+
+        // Assert:
+        expect(component.editing).toBe(false);
+        expect(props.onEdit).not.toHaveBeenCalled();
```
Jaguar0625, line 119: also check component.value

__tests__/components/FormInputEditable.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+        const wrapper = getFormInputEditableWrapper(props);
+        const component = wrapper.vm as FormInputEditableTs;
+
+        // Act:
+        await wrapper.find('button').trigger('click');
+        await wrapper.find('input').setValue(newAccountName);
+        component.finishEdition();
+
+        // Assert:
+        expect(component.value).toBe(newAccountName);
```
Jaguar0625, line 87: also check component.editing

__tests__/components/FormInputEditable.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+        // Act:
+        await wrapper.find('button').trigger('click');
+        await wrapper.find('input').setValue(newAccountName);
+        component.cancelEdition();
+
+        // Assert:
+        expect(component.editing).toBe(false);
+        expect(props.onEdit).not.toHaveBeenCalled();
+    });
+});
```
Jaguar0625, line 121: consider test where component.editing is true

Jaguar0625 (changes requested):

__tests__/components/HardwareConfirmationButton.spec.ts
yilmazbahadir: yea, I used it in other projects. We use jest, not chai here. I already tried something like below but it couldn't catch it. So I was looking at it and that's why I didn't add you as reviewer, since it wasn't complete yet. Thanks for reviewing though. Will address the comments 👍 

```
expect(component.signTransaction(transaction).toPromise()).rejects.toThrow(expectedResult.payload.error);
```

__tests__/components/DurationInput.spec.ts
yilmazbahadir: here is how it's been calculated: https://github.com/symbol/desktop-wallet/blob/48a69cf1c0719cee71f04dbbcf8c5736ee6c9a99/src/core/validation/ValidationRuleset.ts#L47

And this is how it's presented in the UI (both block count and human readable value)
image

I think it's the block count (86400 * 30 secs = 30 days)

__tests__/components/FormInputEditable.spec.ts
yilmazbahadir: it will map to undefined for value prop.

__tests__/components/FormInputEditable.spec.ts
yilmazbahadir: added 👍

__tests__/components/HardwareConfirmationButton.spec.ts
yilmazbahadir: Found the issue and fixed 👍

__tests__/components/ErrorTooltip.spec.ts
yilmazbahadir: can you elaborate on this a little bit more pls? they are expected to be the same.
L70 and L71 =>
https://github.com/symbol/desktop-wallet/blob/5f3f6a485191b76f9b68b35ecdd0ab41fecd0293/__tests__/components/ErrorTooltip.spec.ts#L71

__tests__/components/DurationInput.spec.ts
yilmazbahadir: yes, they are testing; removed the TODO line.

__tests__/components/ErrorTooltip.spec.ts
Jaguar0625: i meant add test that changes to value to 'something' and then to '' or undefined.

__tests__/components/DurationInput.spec.ts
Jaguar0625: do you know where/how the wallet is pulling the configuration from the network?

Jaguar0625: asking because in the config file for the server, the setting looks like this:
```
minNamespaceDuration = 30d
```

__tests__/components/ErrorTooltip.spec.ts
yilmazbahadir: Refactored the component so that it can show multiple errors as a concatenated string.
 For given errors:['Error 1', 'Error 2', '', ' '] it will filter out the empty strings and present the following:
* Error 1
* Error 2

__tests__/components/DurationInput.spec.ts
yilmazbahadir: Sorry, somehow I missed this comment.

I think it's here, getting from the network/properties endpoint:
https://github.com/symbol/desktop-wallet/blob/48a69cf1c0719cee71f04dbbcf8c5736ee6c9a99/src/services/NetworkService.ts#L77

And merging with the defaults:
https://github.com/symbol/desktop-wallet/blob/48a69cf1c0719cee71f04dbbcf8c5736ee6c9a99/src/services/NetworkService.ts#L133

And yes, you're right, it's returning 30d from the endpoint 🤔 
I will look into it deeper tomorrow.

__tests__/components/ErrorTooltip.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+    });
+
+    test('displayed error returns errors as concatenated string', () => {
+        // Arrange:
+        const errors = ['error 1', 'error 2'];
+        const wrapper = getErrorTooltipWrapper(errors);
+        const component = wrapper.vm as ErrorTooltipTs;
+
+        // Act + Assert:
+        expect(component.displayedError).toBe(errors.map((err) => `* ${err}`).join('\n'));
```
Jaguar0625, line 69: i think it's clearer for reader of test to check against constant string i.e. '* error 1\n* error 2`

__tests__/components/ErrorTooltip.spec.ts
```diff
@@ -0,0 +1,96 @@
...
+        expect(component.errored).toBe(true);
+    });
+
+    test('displayed error filters errors that are empty strings', () => {
+        // Arrange:
+        const nonEmptyErrors = ['error 1', 'error 2'];
+        const emptyErrors = ['', ' '];
```
Jaguar0625, line 76: since you're using trim, you should include some whitespace only strings too

__tests__/components/DurationInput.spec.ts
yilmazbahadir: When I checked it, I found out that `30d` is being converted to seconds in:
https://github.com/symbol/desktop-wallet/blob/af3c661ed6c921b2528dc55cfb1c16ad13ff6dab/src/core/utils/NetworkConfigurationHelpers.ts#L160

So both the calculation and the presentation seem to be fine (also had @cryptoBeliever looked at it). There was one issue that the validation was always using `testnet` configurations, I fixed that issue and updated the tests accordingly.

__tests__/components/ErrorTooltip.spec.ts
yilmazbahadir: fixed

__tests__/components/ErrorTooltip.spec.ts
yilmazbahadir: fixed

PR #1889, opened 2022-08-11
[wallet/desktop] feat: add multisig toggle to allow/disallow signing multisig transactions from unknown accounts

OlegMakarenko:
## Problem

The current set of measures is insufficient to protect users from multi-signature scam transactions.

## Current behaviour

Wallet presents a cosignature form explaining the risks of signing multi-signature transactions from unknown accounts.

## Solution

- Disallow signing multisig transactions from unknown accounts by default.
- Present warning instead.
- Added the advanced settings section with "toggle" (dropdown) which can turn on/off this security protection.

image
image


cryptoBeliever (dismissed):

__tests__/views/modals/ModalTransactionCosignature.spec.ts
```diff
@@ -244,4 +263,132 @@ describe('ModalTransactionCosignature', () => {
...
+
+        test('positive when signer is a member of multisig tree', async () => {
+            // Arrange
+            const addressBook = new AddressBook([]);
+            const isAddressInMultisigTree = true;
+            const expectAlertToBeShown = true;
+
+            // Act + Assert:
+            await runKnownSignerAlertTest(addressBook, isAddressInMultisigTree, expectAlertToBeShown);
+        });
```
Jaguar0625, line 365: question: why this behavior?

OlegMakarenko: If the account is in an address book or a member of the multisig tree, we consider that account to be trusted (or known). In that case we do not present the unknown signer **warning** panel (the puprle one). Instead we present a sign form with a small **alert** box which ask user to verify transaction before signing.

__tests__/views/forms/FormAdvancedSettings.spec.ts
```diff
@@ -0,0 +1,224 @@
...
+            await component.$nextTick();
+            const buttonElement = wrapper.find(buttonSelector);
+
+            // Assert:
+            expect(component.isSubmitDisabled).toBe(false);
+            expect(buttonElement.attributes('disabled')).toBeFalsy();
+        });
+    });
+
+    describe('reset()', () => {
```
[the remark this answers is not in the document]
OlegMakarenko: Renamed method to `refresh`.

PR #1903, opened 2022-08-24
[wallet/desktop] feat: one step harvesting impl and unit test improvements

yilmazbahadir:
## What was the issue?
- Delegated Harvesting setup process was 2 step. https://github.com/symbol/desktop-wallet/issues/1820
- Multisig account dlg harvesting was failing sometimes.

## What's the fix?
- Merged key link and activation steps into one step.
- Fixed the multisig account issues
- Fixed status inconsistencies
- Added refresh button for the harvesting status and stats
- Improved the texts and the auto status refresh logic
- A bit of refactoring
- Integrated Testing Library 
- Added unit test for HarvestStatisticsPanel component

Notes:
1. Unit tests for the FormPersistentDelegationRequestTransaction component will be added in another PR.
2. Following keys in the language files needs a translation: 
   - starting
   - harvesting_keys_linked_next_step_guide
   - activate_delegated_harvesting_message
3. Also added the unit test for ImportanceScoreDisplay component in this PR (which can be removed if requested)


cryptoBeliever:
@yilmazbahadir I got a few questions/improvements:

# There is possible to have inconsistent VRF/Remote private keys.

Steps:
## start harvesting for account X in wallet A
## stop and start harvesting for account X in wallet B

Result: In the wallet, A is present correct VRF, Remote public keys. But when you click "show" for a private key we can see an old invalid private key. This is causing the invalid persistent delegated message to be constructed.

Expected result: I think we should detect that corresponding private keys are not matching the public keys in the network. Maybe when are we decrypt keys during starting harvesting? If saved private keys are invalid unlink and remove also VRF and Remote private keys.

# When are saved keys (remote, VRF) for multi-sig accounts? I have the following scenario:

I have multi-sig M1 with cosigners C1 and C2
## Start harvesting for multi-sig account M1 from cosigner C1
## Go to cosigner C2, sign transaction, and wait for confirmation (wait on account view without going back to harvesting view)

After confirmation, I have no private keys (remote and VRF have just public keys) saved for multisig M1 when I'm checking using C1 or C2.

# For Ledger integration we asked twice for a password when activating delegated harvesting. Is it needed?

## First is presented window with text "To activate delegated harvesting, we need your password to decrypt your existing Remote Account and VRF keys."
## Next "As a Ledger user your password is needed to encrypt your remote account and VRF private keys"

Video: https://www.loom.com/share/b1fc4b244e3e48289271140b569d2ae8

cryptoBeliever:
All those https://github.com/symbol/desktop-wallet/pull/1903#issuecomment-1231853733 (and more) fixed and looks good.
I've tested:
- simple account
- multi-sig account
- ledger wallet simple account
- ledger wallet multi-sig account
For all:
- start harvesting
- start harvesting when we have already keys linked
- start harvesting when we have keys linked but we don't have keys (vrf, remote) locally
- start harvesting when we have keys linked and we have keys locally which does not match linked public keys
- start harvesting when we have keys linked and matching private keys locally
- stop harvesting

Node operator case:
- automatically detect harvesting status
- stop harvesting should be hidden if the status active

cryptoBeliever (dismissed):

cryptoBeliever (approved):

__tests__/components/HarvestStatisticsPanel.spec.ts
```diff
@@ -0,0 +1,299 @@
...
+            if (type === 'harvesting/FETCH_STATUS') {
+                store.commit('harvesting/status', HarvestingStatus.ACTIVE);
+            }
+        });
+
+        // Act:
+        await Vue.nextTick();
+        jest.advanceTimersByTime(HARVESTING_STATUS_POLLING_INTERVAL_SECS * 1000);
+        await Vue.nextTick();
+        '';
```
Jaguar0625, line 195: delete?

__tests__/components/HarvestStatisticsPanel.spec.ts
```diff
@@ -0,0 +1,299 @@
...
+        await Vue.nextTick();
+        jest.advanceTimersByTime(HARVESTING_STATUS_POLLING_INTERVAL_SECS * 1000);
+        store.commit('harvesting/status', HarvestingStatus.FAILED);
+        await Vue.nextTick();
+
+        // Assert:
+        expect(getByTestId('status-text').textContent).toBe(i18n.t(`harvesting_status_failed`).toString());
+    });
+
+    test(`refresh status blocks periodically when harvesting status is active`, async () => {
```
Jaguar0625, line 246: since you're checking refresh, you might want to check that initial value is set correctly

src/store/Harvesting.ts
```diff
@@ -152,14 +156,25 @@ export default {
             commit('setPollingTrials', pollingTrials);
         },
         async FETCH_STATUS({ commit, rootGetters, dispatch }, node?: [string, NodeModel]) {
-            const currentSignerAccountInfo: AccountInfo = rootGetters['account/currentSignerAccountInfo'];
+            const csAccountInfo: AccountInfo = rootGetters['account/currentSignerAccountInfo'];
```
Jaguar0625, line 159: i'd suggest not shortening and using original name

__tests__/components/HarvestStatisticsPanel.spec.ts
```diff
@@ -0,0 +1,299 @@
...
+    });
+
+    Object.keys(HarvestingStatus).map((key) => {
+        test(`harvesting status is changed to ${key} when refresh button is clicked`, async () => {
+            // Arrange:
+            const dispatch = jest.fn();
+            const { getByTestId, getByRole, store } = renderComponent({}, dispatch);
+
+            dispatch.mockImplementationOnce((type: string) => {
+                if (type === 'harvesting/FETCH_STATUS') {
```
Jaguar0625, line 137: aren't you explicitly changing the status here?
it seems like you're hooking this event and then forwarding to store?
can we not use the "real" dispatcher here?

src/views/forms/FormPersistentDelegationRequestTransaction/FormPersistentDelegationRequestTransactionTs.ts
```diff
@@ -92,10 +92,9 @@ import { feesConfig as defaultFeesConfig, networkConfig } from '@/config';
 export enum HarvestingAction {
     START = 1,
     STOP = 2,
-    SWAP = 3,
-    ACTIVATE = 4,
-    SINGLE_KEY = 5,
+    SINGLE_KEY = 3,
```
Jaguar0625, line 95: are you missing tests for these changes?

src/views/forms/FormPersistentDelegationRequestTransaction/FormPersistentDelegationRequestTransactionTs.ts
```diff
@@ -381,20 +367,45 @@ export class FormPersistentDelegationRequestTransactionTs extends FormTransactio
...
+                txs.push(accountUnlinkTx);
+            }
+            if (this.isVrfKeyLinked && !this.vrfPrivateKey) {
+                const vrfUnlinkTx = this.createVrfKeyLinkTx(
+                    this.currentSignerAccountInfo.supplementalPublicKeys?.vrf.publicKey,
+                    LinkAction.Unlink,
+                    maxFee,
+                );
+                txs.push(vrfUnlinkTx);
+            }
```
Jaguar0625, line 386: why is START unlinking? to clear existing values? in that case won't it complete with nothing being linked?

src/views/forms/FormPersistentDelegationRequestTransaction/FormPersistentDelegationRequestTransactionTs.ts
```diff
@@ -381,20 +367,45 @@ export class FormPersistentDelegationRequestTransactionTs extends FormTransactio
...
                 txs.push(vrfKeyLinkTx);
             }
 
-            if (!this.isNodeKeyLinked) {
-                const nodeLinkTx = this.createNodeKeyLinkTx(this.formItems.nodeModel.nodePublicKey, LinkAction.Link, maxFee);
-                txs.push(nodeLinkTx);
-            }
+            // link node key
+            const nodeLinkTx = this.createNodeKeyLinkTx(this.formItems.nodeModel.nodePublicKey, LinkAction.Link, maxFee);
+            txs.push(nodeLinkTx);
```
Jaguar0625, line 398: while this is required for delegated harvesting, it is not required for non-delegated harvesting. 
whereas the other two links (remote, vrf) are required for all forms of harvesting.
unsure if this should be mandatory?

also, unlike the others, it appears this link is not cleared? is that intentional?

Jaguar0625 (changes requested):

src/store/Harvesting.ts
yilmazbahadir: original name is being used in the line 168(and in the rest of the function), that's why I changed here. In line 168 we are refreshing the current signer account info object by a remote call. Previously I tried to keep the name `currentSignerAccountInfo`  in line 159 and call the new constant `refreshedCurrentSignerAccountInfo`, but didn't feel right when I looked at it so I decided to go with this version. I'd like to hear if you have any suggestions on naming here?

src/views/forms/FormPersistentDelegationRequestTransaction/FormPersistentDelegationRequestTransactionTs.ts
yilmazbahadir: Yes, it will be handled in the following PRs. I had attached the following note in the description.
> Unit tests for the FormPersistentDelegationRequestTransaction component will be added in another PR.

src/views/forms/FormPersistentDelegationRequestTransaction/FormPersistentDelegationRequestTransactionTs.ts
yilmazbahadir: pls see https://github.com/symbol/desktop-wallet/pull/1862#issuecomment-1136915774

1) Start harvesting:

Start harvesting should send one aggregate transaction containing:

- unlink remote account (only if an account is already linked and we don't have a private key for it in storage)
- unlink VRF key (only if an account is already linked and we don't have a private key for it in storage)
- unlink node key (if a node key is linked - we need to unlink the old one)
- link remote account (only if we had to remove the old one - then we have to generate and link a new key)
- link VRF key (only if we had to remove the old one - then we have to generate and link the new key)
- link node key (always)
- persistent delegated message (always)

2) Stop harvesting: 

Start harvesting should send one aggregate transaction containing:
- unlink node key (only if a node key is linked do we need to unlink the old one)
- unlink remote account (only if an account is already linked and we don't have the private key for it in storage)
- unlink VRF key (only if an account is already linked and we don't have the private key for it in storage)

src/views/forms/FormPersistentDelegationRequestTransaction/FormPersistentDelegationRequestTransactionTs.ts
yilmazbahadir: > while this is required for delegated harvesting, it is not required for non-delegated harvesting.

this form is handling `delegated-harvesting` only, `non-delegating` harvesters (node-owners) can only monitor their state here.

> also, unlike the others, it appears this link is not cleared? is that intentional?

if it's already linked, it's always being unlinked first(on line 361). 
so if linked regardless of the action (STOP or START)  => unlink the current node key link
and if the action is START, additionally => link the new node key

__tests__/components/HarvestStatisticsPanel.spec.ts
yilmazbahadir: we are mocking the vuex store in our testing where the vuex is involved. And this is part of it.
when the component dispatches `harvesting/FETCH_STATUS` event, it's expected to fetch(remote call) , calculate the new status and commit to the store.
since we want to mock this behaviour, after the first render before the refresh button is clicked we are setting the scene, so that the refresh event can set the correct status.

> can we not use the "real" dispatcher here?

I'm afraid not.

src/store/Harvesting.ts
gimre-xymcity: naming is fine, if there's line limit hit, can't it be split in half like
```ts
const currentSignerAccountInfo: AccountInfo =
[tab][tab]rootGetters['account/currentSignerAccountInfo'];
```

yilmazbahadir: thanks @gimre-xymcity, there wasn't any line limit issue but refactored to an Address type instead as it will clear the confusion and fit better 👍

Jaguar0625 (approved):

PR #1906, opened 2022-08-28
[wallet/desktop] feat: unit test spinner

AnthonyLaw:
## What was the issue?
- missing unit test in the `SpinnerLoading` component.

## What's the fix?
- Added unit test  `SpinnerLoading` component.
- Removed unused components `SpinnerDots`


__tests__/components/SpinnerLoading.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+            wrapper.vm.closeScreen();
+
+            // Assert:
+            expect(wrapper.vm.$store.dispatch).toBeCalledWith('app/SET_LOADING_OVERLAY', {
+                show: false,
+                message: '',
+            });
+        });
+    });
+});
```
Jaguar0625, line 64: can you add test for what is rendered when disableCloseButton is true and false?

AnthonyLaw: I was not able to test it, it just render out `<div />` as output.
I will try again to play around.

PR #1910, opened 2022-09-02
[wallet/desktop] feat: unit test signer components

AnthonyLaw:
## What was the issue?
- missing unit test in the `signer` related component.

## What's the fix?
- Added unit test  `signerSelector`, `signerBaseFilter` and `signerList`  component.

### DIFF
--- a/__tests__/components/SignerBaseFilter.spec.ts
+++ b/__tests__/components/SignerBaseFilter.spec.ts
@@ -61,7 +61,7 @@ describe('components/SignerBaseFilter', () => {
         );
 
     describe('component', () => {
-        const runBasicSelectedSignerInMountTests = (props, expectResult) => {
+        const runBasicSelectedSignerInMountTests = (props, expectedResult) => {
             // Arrange:
             const wrapper = getSignerBaseFilterWrapper({
                 rootSigner: mockSigner,
@@ -71,23 +71,21 @@ describe('components/SignerBaseFilter', () => {
             const vm = wrapper.vm as SignerBaseFilterTs;
 
             // Act + Assert:
-            expect(vm.selectedSigner).toBe(expectResult);
+            expect(vm.selectedSigner).toBe(expectedResult);
         };
 
         test('renders only account in the list when parent signer does not exist', () => {
             // Arrange:
+            const { parentSigners, ...rest } = mockSigner;
+
             const wrapper = getSignerBaseFilterWrapper({
-                rootSigner: {
-                    label: 'wallet 1',
-                    address: simpleWallet1.address,
-                    multisig: false,
-                    requiredCosigApproval: 0,
-                    requiredCosigRemoval: 0,
-                },
+                rootSigner: rest,
             });
 
-            // Act + Assert:
+            // Act:
             const options = wrapper.findAll('option');
+
+            // Assert:
             expect(options.length).toBe(1);
             expect(options.at(0).text()).toBe('wallet 1');
             expect(options.at(0).attributes().value).toBe(simpleWallet1.address.plain());
@@ -118,19 +116,29 @@ describe('components/SignerBaseFilter', () => {
             expect(options.at(1).attributes().value).toBe(simpleWallet2.address.plain());
         });
 
-        test('created set selected signer to chosen signer when isAggregate is true and chosen signer does not empty', () => {
+        test('selects specified signer when isAggregate is true when specified signer is not empty', () => {
             const chosenSigner = simpleWallet2.address.plain();
             runBasicSelectedSignerInMountTests({ isAggregate: true, chosenSigner }, chosenSigner);
         });
 
-        test('created set selected signer to current signer when isAggregate is false and chosen signer is empty', () => {
+        test('selects specified signer when isAggregate is true when specified signer is empty', () => {
             const chosenSigner = '';
+            runBasicSelectedSignerInMountTests({ isAggregate: true, chosenSigner }, simpleWallet1.address.plain());
+        });
+
+        test('selects specified signer when isAggregate is false when specified signer is empty', () => {
+            const chosenSigner = '';
+            runBasicSelectedSignerInMountTests({ isAggregate: false, chosenSigner }, simpleWallet1.address.plain());
+        });
+
+        test('selects specified signer when isAggregate is false when specified signer is not empty', () => {
+            const chosenSigner = simpleWallet2.address.plain();
             runBasicSelectedSignerInMountTests({ isAggregate: false, chosenSigner }, simpleWallet1.address.plain());
         });
     });
 
     describe('onSignerChange', () => {
-        test('emit selected signer value when click on first option in selector list', () => {
+        test('emits selected signer value when click on any options in selector list', () => {
             // Arrange:
             const wrapper = getSignerBaseFilterWrapper({
                 rootSigner: mockSigner,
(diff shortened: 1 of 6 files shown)


__tests__/components/SignerBaseFilter.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+                ...props,
+            },
+            {
+                OptionGroup: true,
+                Icon: true,
+            },
+        );
+
+    describe('component', () => {
+        const runBasicSelectedSignerInMountTests = (props, expectResult) => {
```
Jaguar0625, line 64: expect*ed

__tests__/components/SignerBaseFilter.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+        test('renders only account in the list when parent signer does not exist', () => {
+            // Arrange:
+            const wrapper = getSignerBaseFilterWrapper({
+                rootSigner: {
+                    label: 'wallet 1',
+                    address: simpleWallet1.address,
+                    multisig: false,
+                    requiredCosigApproval: 0,
+                    requiredCosigRemoval: 0,
+                },
```
Jaguar0625, line 86: isn't this mockSigner?

__tests__/components/SignerBaseFilter.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+            });
+
+            // Act + Assert:
+            const options = wrapper.findAll('option');
```
Jaguar0625, line 90: findAll should be // Act:
rest should be // Assert:

__tests__/components/SignerBaseFilter.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+            expect(options.length).toBe(2);
+            expect(options.at(0).text()).toBe('wallet 1');
+            expect(options.at(0).attributes().value).toBe(simpleWallet1.address.plain());
+            expect(options.at(1).text()).toBe('(Multisig)wallet 2');
+            expect(options.at(1).attributes().value).toBe(simpleWallet2.address.plain());
+        });
+
+        test('created set selected signer to chosen signer when isAggregate is true and chosen signer does not empty', () => {
```
Jaguar0625, line 121: selects specified signer when isAggregate is true when specified signer is not empty

__tests__/components/SignerBaseFilter.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+            const chosenSigner = simpleWallet2.address.plain();
+            runBasicSelectedSignerInMountTests({ isAggregate: true, chosenSigner }, chosenSigner);
+        });
+
+        test('created set selected signer to current signer when isAggregate is false and chosen signer is empty', () => {
+            const chosenSigner = '';
+            runBasicSelectedSignerInMountTests({ isAggregate: false, chosenSigner }, simpleWallet1.address.plain());
+        });
```
Jaguar0625, line 129: should you test the two other combinations too for completeness?

__tests__/components/SignerBaseFilter.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+    });
+
+    describe('onSignerChange', () => {
+        test('emit selected signer value when click on first option in selector list', () => {
```
Jaguar0625, line 133: emit*s

__tests__/components/SignerBaseFilter.spec.ts
```diff
@@ -0,0 +1,165 @@
...
+            // Act:
+            wrapper.findAll('option').at(0).trigger('input');
+
+            // Assert:
+            expect(wrapper.emitted('signer-change')).toEqual([[simpleWallet1.address.plain()]]);
+        });
+    });
+
+    describe('onCurrentSignerChange', () => {
+        test('set selected signer to current signer', () => {
```
Jaguar0625, line 148: set*s

__tests__/components/SignerBaseFilter.spec.ts
Jaguar0625: why are you specifying 'click on first' isn't it click on any?

__tests__/components/SignerSelector.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+                            },
+                        ],
+                    },
+                });
+
+                // Act + Assert:
+                expect(wrapper.findComponent(SignerBaseFilter).exists()).toBe(true);
+            });
+
+            test('hide signer base filter component when parent signer does not exist', () => {
```
Jaguar0625, line 66: hide*s

__tests__/components/SignerSelector.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+                // Arrange:
+                const wrapper = getSignerSelectorWrapper();
+
+                // Act + Assert:
+                expect(wrapper.findComponent(SignerBaseFilter).exists()).toBe(false);
+            });
+
+            test('display label when root signer is multisig and parent signer does not exist', () => {
```
Jaguar0625, line 74: display*s

__tests__/components/SignerSelector.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+                value: simpleWallet1.address.plain(),
+            });
+
+            const vm = wrapper.vm as SignerSelectorTs;
+
+            // Act + Assert:
+            expect(vm.chosenSigner).toEqual(simpleWallet1.address.plain());
+        });
+
+        test('set new chosen signer when new signer provided', () => {
```
Jaguar0625, line 106: set*s

__tests__/components/SignerSelector.spec.ts
```diff
@@ -0,0 +1,121 @@
...
+
+            const wrapper = getSignerSelectorWrapper();
+
+            const vm = wrapper.vm as SignerSelectorTs;
+
+            // Act:
+            vm.onSignerSelectorChange(newSigner);
+
+            // Assert:
+            expect(wrapper.emitted('input')).toEqual([[newSigner]]);
```
Jaguar0625, line 118: should you check vm.chosenSigner too ?

__tests__/components/SignerBaseFilter.spec.ts
AnthonyLaw: yes, but the main difference is that `parantSigners` do not exist.
but I find a better way to write it.

`const { parentSigners, ...rest } = mockSigner;`

__tests__/components/SignerBaseFilter.spec.ts
AnthonyLaw: That is because I triggered the first option in the test.
but yeah, it makes sense I will update to click on any

__tests__/components/SignerBaseFilter.spec.ts
AnthonyLaw: agree, I will update 👍🏼

__tests__/components/SignerSelector.spec.ts
AnthonyLaw: this one is interesting,  the `vm.chosenSigner` is not updated in the unit test, but it works in production.
I will dig more on this.

AnthonyLaw: After doing some digging, and discussing with the team. We are safe to **skip** `check vm.chosenSigner` in this case.

Vue `$emit` event is used to passing date from `child` to `parent`. So in this case `wrapper.emitted('input')` is enough to verify.

Jaguar0625 (approved):

PR #1912, opened 2022-09-13
tests/harvested blocklist page

yilmazbahadir:
* Added unit test for harvested block list page

### DIFF
--- a/__tests__/views/pages/harvesting/HarvestedBlocksList.spec.ts
+++ b/__tests__/views/pages/harvesting/HarvestedBlocksList.spec.ts
@@ -1,0 +1,177 @@
+/*
+ * (C) Symbol Contributors 2022
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


__tests__/views/pages/harvesting/HarvestedBlocksList.spec.ts
```diff
@@ -0,0 +1,177 @@
...
+            }
+        });
+
+        // Act:
+        getByTestId('btn-next-page').click();
+        await Vue.nextTick();
+
+        // // Assert:
+        expect(queryByText('1000')).toBeNull();
+        expect(getByText('21000')).toBeDefined();
```
Jaguar0625, line 146: is it possible to search element by class or id?

yilmazbahadir: using css selectors or IDs are kind of anti-pattern in testing library philosophy (event test-id's are considered as escape hatch), in this case since we use iview library's Table component, we couldn't add any test-id's so went with the text query
> One of the guiding principles of the Testing Library APIs is that they should enable you to test your app the way your users use it

Jaguar0625 (approved):

PR #1913, opened 2022-09-14
[wallet/desktop] feat: rental fee component unit test

AnthonyLaw:
## What was the issue?
- missing the `RentalFee` component unit test.

## What's the fix?
- Added unit test  `RentalFee` component.
- minor refactor.

### DIFF
--- a/__tests__/components/RentalFee.spec.ts
+++ b/__tests__/components/RentalFee.spec.ts
@@ -17,7 +17,7 @@ import RentalFee from '@/components/RentalFees/RentalFee.vue';
 import { render } from '@testing-library/vue';
 import { getStore } from '@MOCKS/Store';
 import i18n from '@/language/index';
-import { mosaicsMock } from '@MOCKS/mosaics';
+import { networkMock } from '@MOCKS/network';
 import { UInt64 } from 'symbol-sdk';
 
 describe('components/RentalFee', () => {
@@ -35,7 +35,7 @@ describe('components/RentalFee', () => {
         const mockMosaicStore = {
             namespaced: true,
             state: {
-                networkCurrency: mosaicsMock[0],
+                networkCurrency: networkMock.currency,
             },
             getters: {
                 networkCurrency: (state) => state.networkCurrency,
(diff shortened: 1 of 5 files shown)


__tests__/components/RentalFee.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+            duration: 100000,
+        });
+
+        // Act + Assert:
+        expect(getByText('Estimated rental fee:')).toBeDefined();
+        expect(getByTestId('fees').textContent.trim()).toContain(expectedResult);
+    };
+
+    it('renders default estimated rental fee when rentalFeeEstimation does not exist', () => {
+        runBasicEstimatedRentalFee({}, 'mosaic', '0');
```
Jaguar0625, line 88: why are you checking the rendered text vs the result of rentalFeeAmount ?

src/components/RentalFees/RentalFeeTs.ts
```diff
@@ -32,18 +32,17 @@ export class RentalFeeTs extends Vue {
         if (this.rentalEstimation) {
```
Jaguar0625, line 32: when will this be unset?

AnthonyLaw: https://github.com/symbol/desktop-wallet/blob/7b5f35bf899d1f1e60d33fd4daab5ba3d684485e/src/store/Network.ts#L556-L563

hmm, I think most properly it won't be unset.
because when the network/node is down, the wallet won't work.
if the network is in working condition, it should always receive rental fees.

I will remove this `if` condition.

__tests__/components/RentalFee.spec.ts
AnthonyLaw: Recently we had a discussion in our daily meeting, about using `@testing-library/vue`
The test should focus more on user behavior or experience in the wallet.

so **rendered text** will be more accurate because it will display to the user.

What do you think about it?

Jaguar0625: can you please summarize this discussion / decision in a discord channel for further discussion?

Jaguar0625 (approved):

PR #1914, opened 2022-09-26
tests/delegated harvesting form

yilmazbahadir:
## What was the issue?
- unit tests were missing for Persistent delegated harvesting and Transfer transaction forms
- was mostly testing the component internals without worrying about the user interaction, this is fine and also necessary for the _basic component_ testing but for the complex pages we need to simulate the user interaction in order to make sure the user experience remains intact even the internal logic changes [more info on Testing Library](https://testing-library.com/docs/)

## What's the fix?
- using Testing Library ([@testing-library/vue](https://testing-library.com/docs/vue-testing-library/intro/)) for the view/page testing
    - a wrapper library around the `vue-test-utils` library that we use in our component testing
    - mounting components the way they are presented to the user in the browser environment
    - interacting with the view/page exactly the same way  the user would do (finding DOM elements by roles, texts etc. and clicking, typing like a user)
- used real vuex store (exactly the one being used when the application starts)
    - refactored the AppStore and the code so that the same store can be used in isolation in every test
- used mock http server ([msw](https://www.npmjs.com/package/msw)) for network request/response mocking
    - created `__mock__/Http.ts` for the basic responses for the store initialization for the accounts(`__mocks__/Account.ts`) used in every test
    - since we are mocking http responses instead of mocking `symbol-sdk` methods, this will reduce the level of refactoring needed during the new SDK migration
- added unit testing utility `TestUIHelpers.ts` for the transactional form testing (Harvesting, Transfer, Multisig Account, Namespace, Mosaic etc.)
    - In order to test the utility, added a basic unit test for the Transfer Transaction Form which  is covering ~70% already
- added user-centric unit test for Delegated Harvesting Form


gimre-xymcity (approved):

__mocks__/Accounts.ts
```diff
@@ -51,9 +51,9 @@ export const simpleWallet2 = SimpleWallet.createFromPrivateKey(
 );
 
 export const WalletsModel1: AccountModel = {
-    id: 'someId',
+    id: 'WalletsModel1',
     node: '',
-    profileName: 'profile_name',
+    profileName: 'profile1',
```
0x6861746366574, line 56: I'd prefer that we have clearer naming throughout. 'WalletsModel2' and 'Profile1' is the naming equivalent of 'Final Edit_final_final_v2.docx'.

__mocks__/profiles.ts
```diff
@@ -20,10 +20,11 @@ const TEST_PROFILES = {
     profile1: {
         generationHash: 'ACECD90E7B248E012803228ADB4424F0D966D24149B72E58987D2BF2F2AF03C4',
         profileName: 'profile1',
-        hint: 'password is password',
+        hint: 'password is Password1',
```
0x6861746366574, line 23: ..lets, uh, not do 'password1'. No, it doesn't need to be secure, but be a bit creative in your naming so you don't overlap with other products. 

You could have chosen rum; oxford; billboard; lotus...any password and name combination!

0x6861746366574 (commented):
Good work, @yilmazbahadir. I want better naming schemas for our functions, modals, variables, etc. across the board in future products (no 'password1', and 'Account2' - instead, more clear naming such as ParentKey and ChildKey, UserAccount_Generated, UserAccount_Child, or even UserAccountOne) but otherwise, looks good.

__mocks__/Accounts.ts
yilmazbahadir: these names were inherited so didn't want to change, you're right, noted 👍

__mocks__/profiles.ts
yilmazbahadir: `Password1` is the improved version over old `password` :) picked a simple password to show that it requires a capital and a numeric char.

PR #1919, opened 2022-10-21
[wallet/desktop] feat: upload-qr component unit test

AnthonyLaw:
## What was the issue?
- missing unit test in the `UploadQRCode` component.

## What's the fix?
- Added unit test  `UploadQRCode` component.

### DIFF
--- a/__tests__/components/QRCode/UploadQRCode.spec.ts
+++ b/__tests__/components/QRCode/UploadQRCode.spec.ts
@@ -20,6 +20,9 @@ import { QRCodeType } from 'symbol-qr-library';
 import { networkMock } from '@MOCKS/network';
 
 describe('components/QRCode/UploadQRCode', () => {
+    beforeEach(() => {
+        URL.createObjectURL = jest.fn();
+    });
     const getUploadQRCodeWrapper = (state = {}, props = {}) => {
         const networkStore = {
             namespaced: true,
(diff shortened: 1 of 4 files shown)


__tests__/components/QRCode/UploadQRCode.spec.ts
```diff
@@ -0,0 +1,183 @@
...
+
+            // Assert:
+            expect(vm.invalidType).toBe(false);
+            expect(vm.imageFileName).toBe('invoice_qr.png');
+            expect(vm.$refs.qrcodeCapture.onChangeInput).toHaveBeenCalledWith({
+                target: {
+                    files: [file],
+                },
+            });
+            expect(fileReaderSpy).toHaveBeenCalledWith(file);
```
Jaguar0625, line 180: check vm.image and return code

setupTests.js
```diff
@@ -0,0 +1,3 @@
+console.log('Setting up global test stubs...');
+
+URL.createObjectURL = jest.fn();
```
Jaguar0625, line 3: why do you need this?

AnthonyLaw: Initially rendering the component prompts the error `URL.createObjectURL is not a function`.
I believe some dependency is required `createObjectURL` method. So I just mock it into global jest.

__tests__/components/QRCode/UploadQRCode.spec.ts
AnthonyLaw: that is an interesting part, the `fileReader.onload` is not triggered in my unit test.
I will check on this problem.

setupTests.js
Jaguar0625: why global instead of just for the specific test(s) ?

AnthonyLaw: I did try to move the specific test, but it won't work. 

I found some good explanations here.
https://github.com/mapbox/mapbox-gl-js/issues/9889#issuecomment-664633443

Jaguar0625 (approved):

PR #1924, opened 2022-11-03
[wallet/desktop] fix: symbol-sdk transaction info fixes in tests

yilmazbahadir:
## What was the issue?
After symbol-sdk@2.0.3 upgrade some of the unit tests began to fail due to the constructor signature change in TransactionInfo class.

## What's the fix?
Fixed the parameters need to be passed to TransactionInfo constructor.

### DIFF
--- a/__mocks__/transactions.ts
+++ b/__mocks__/transactions.ts
@@ -25,6 +25,8 @@ export const transferTransaction = new TransferTransaction(
         UInt64.fromUint(108907),
         0,
         '',
+        UInt64.fromUint(1566230400),
+        100,
         '9661088F22C2DC72E76EE2B0F07BFC0D8E98EEE0CC8AE274362EDBB44F164F16',
         '9661088F22C2DC72E76EE2B0F07BFC0D8E98EEE0CC8AE274362EDBB44F164F16',
     ),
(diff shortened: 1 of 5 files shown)


cryptoBeliever:
@yilmazbahadir 
- when I run `npm run build` I got no test errors :heavy_check_mark: 
- when I run `npm run test` I got 4 tests failing (problem with localization):

image

- I got also (when running npm run build) many Vue errors. Not sure if this is something to ignore or not.

AnthonyLaw (approved):
All good. 
IMO but instead of `undefined` we can just randomly mock the value.
because seems all mocking tx is confirmed. So `feeMultiplier` and `timestamp` should not `undefined`

Jaguar0625 (changes requested):
> IMO but instead of undefined we can just randomly mock the value.
because since all mocking tx is confirmed. So feeMultiplier and timestamp should not undefined

I agree with @AnthonyLaw. Please update.

yilmazbahadir:
> > IMO but instead of undefined we can just randomly mock the value.
> > because since all mocking tx is confirmed. So feeMultiplier and timestamp should not undefined
> 
> I agree with @AnthonyLaw. Please update.

Updated with mock values.

Jaguar0625 (approved):

PR #1928, opened 2022-11-08
[wallet/desktop] build: fix dependencies issue

AnthonyLaw:
Problem: Electron is incompatible with Ledger library, causing the wallet to be unable to detect Ledger wallet.
Solution: Find the right version of the electron


Wayonb (approved):
👍🏾

PR #2022, opened 2025-08-14
[wallet/desktop]: migrate to node watch service

AnthonyLaw:
Problem: The Statistic service is going to sunset, and it requires migration to the node watch service.
Solution: added nodewatch service

### DIFF
--- a/src/services/NodeService.ts
+++ b/src/services/NodeService.ts
@@ -73,16 +73,11 @@ export class NodeService {
             .then((val) => (statisticsNodes && statisticsNodes.length ? _.uniqBy(val.concat([...statisticsNodes]), 'url') : val));
     }
 
-    public async getNodesFromNodeWatchService(
-        networkType: NetworkType,
-        limit = 30,
-        sslOnly = true,
-        isOffline?: boolean,
-    ): Promise<NodeModel[]> {
+    public async getNodesFromNodeWatchService(networkType: NetworkType, limit = 30, isOffline?: boolean): Promise<NodeModel[]> {
         if (!isOffline && navigator.onLine) {
             try {
                 const nodeWatchService = new NodeWatchService(networkType);
-                let nodeInfos = await nodeWatchService.getNodes(sslOnly, limit, 'random');
+                let nodeInfos = await nodeWatchService.getNodes(limit, 'random');
 
                 if (!nodeInfos) {
                     return undefined;
--- a/src/services/NodeService.ts
+++ b/src/services/NodeService.ts
@@ -134,8 +134,8 @@ export class NodeService {
 
             return this.getNodeModelByMethod(
                 networkType,
-                async (pKey) => {
-                    return await nodeWatchService.getNodeByMainPublicKey(pKey);
+                async (publicKey) => {
+                    return await nodeWatchService.getNodeByMainPublicKey(publicKey);
                 },
                 publicKey,
             );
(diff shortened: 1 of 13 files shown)


src/services/NodeService.ts
```diff
@@ -139,31 +124,38 @@ export class NodeService {
...
         publicKey: string,
         isOffline?: boolean,
     ): Promise<NodeModel> {
         if (!isOffline && navigator.onLine && publicKey) {
+            const nodeWatchService = new NodeWatchService(networkType);
+
             return this.getNodeModelByMethod(
                 networkType,
-                (pKey) => this.createStatisticServiceRestClient(networkConfig[networkType].statisticServiceUrl).getNode(pKey),
+                async (pKey) => {
```
Wayonb: pKey=> key or publicKey

src/services/NodeService.ts
```diff
@@ -139,31 +124,38 @@ export class NodeService {
...
         isOffline?: boolean,
     ): Promise<NodeModel> {
         if (!isOffline && navigator.onLine && nodePublicKey) {
+            const nodeWatchService = new NodeWatchService(networkType);
+
             return this.getNodeModelByMethod(
                 networkType,
-                (npKey) =>
-                    this.createStatisticServiceRestClient(networkConfig[networkType].statisticServiceUrl).getNodeByNodePublicKey(npKey),
+                async (npKey) => {
```
Wayonb: npKey => key or publicKey

src/services/NodeWatchService.ts
```diff
@@ -0,0 +1,74 @@
...
+
+export class NodeWatchService {
+    private readonly url: string;
+
+    constructor(networkType: NetworkType) {
+        this.url = networkConfig[networkType].nodeWatchServiceUrl;
+    }
+
+    private mapResponseToNodeApiInfo(response: any): NodeApiNodeInfo {
+        const endpoint = response.endpoint.replace('http', 'https').replace('3000', '3001');
```
Wayonb: What happens if the is not ssl?

src/services/NodeWatchService.ts
```diff
@@ -0,0 +1,74 @@
...
+            mainPublicKey: response.mainPublicKey,
+            nodePublicKey: response.nodePublicKey,
+            isSslEnabled: response.isSslEnabled,
+            isHealthy: response.isHealthy,
+            restVersion: response.restVersion,
+            wsUrl,
+        };
+    }
+
+    async getNodes(onlySSL = true, limit = 0, order = null): Promise<NodeApiNodeInfo[]> {
```
Wayonb: Does the wallet only support SSL nodes?  If so then should not allow the override of ``onlySSL = true`` .

Wayonb (changes requested):

src/services/NodeWatchService.ts
AnthonyLaw: I assume all nodes will be SSL enabled.

src/services/NodeWatchService.ts
AnthonyLaw: > Does the wallet only support SSL nodes?

I enforce only `https` nodes because some `http` nodes can cause CORS or 301 Moved Permanently errors.

Wayonb (dismissed):
👍🏾

src/services/NodeWatchService.ts
cryptoBeliever: This is likely causing problems when starting harvesting on a non-SSL node.
Steps to reproduce:
1. Start harvesting on an on-SSL node like t.rerena.nemnesia.com
2. Wallet is unable to fetch harvesting status.

The problem is that the wallet searches node by nodePublicKey/mainPublicKey and always uses the endpoint https://<node_domain>:3001, regardless of whether the node supports SSL or not. In my opinion if node does not support SSL we should use http + port 3000.

cryptoBeliever: Fixed.

cryptoBeliever (approved):

PR #2030, opened 2025-08-28
[wallet/desktop] fix: remove manually logic update to https

AnthonyLaw:
Problem: nodewatch will handle the endpoint for https if SSL is enabled.
Solution: Removed manually logic


Wayonb (approved):
👍🏾

